PR #3, opened 2014-07-03
fix flake8 warnings

gurneyalex:


bwrsandman:
I would avoid `# noqa` in __init__.py, it is a bad way of fixing pep8 errors. It would mean to put `# noqa` in every __init__.py

[If you're already doing this](https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_run_flake8), why do you need to add it?

```
    - flake8 . --max-line-length=120 --filename=__init__.py --ignore=F401
    - flake8 . --max-line-length=120 --exclude=__init__.py
```

stock_picking_webkit/__init__.py
```diff
@@ -19,4 +19,4 @@
 #
 ##############################################################################
 
-from . import report
+from . import report  # noqa
```
bwrsandman, line 22: `# noqa` should not be required

yvaucher:
pep8 has been fixed in 7.0

PR #2, opened 2014-07-01
[FIX] replace both Print buttons with the new action

lepistone:

PR #9, opened 2014-10-07
7.0 stock picking webkit

sebastienbeau:
Remove duplicated button for printing picking report.
Since this PR have been merged https://github.com/OCA/stock-logistics-reporting/pull/2 (I agree with it) when the picking is in the done state, odoo will show too button for printing the report (initialy this button print two different report but one is useless). I propose to hide on of this button to have a better experience for the user.

### DIFF
--- a/stock_picking_webkit/stock_view.xml
+++ b/stock_picking_webkit/stock_view.xml
@@ -4,6 +4,17 @@
         <record id="view_picking_out_form" model="ir.ui.view">
             <field name="name">stock.picking.out.form</field>
             <field name="model">stock.picking.out</field>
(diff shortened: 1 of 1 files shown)


eLBati:
Hello @sebastienbeau The purpose of this PR is to never show the 'Print Delivery Slip' button?

yvaucher:
@sebastienbeau If you want to hide this button, wouldn't it be better to specify the states you search in your xpath search and make it invisible eval="1" ?

eLBati:
I think this can be closed

gurneyalex:
@sebastienbeau ping

PR #12, opened 2014-11-30
[ADD] stock_valued: Add amount information to pickings

acysos:
Hello,

Add price, discounts, untaxed amount, taxes and total to picking list.

Greetings

### DIFF
--- a/stock_valued/REAME.rst
+++ b/stock_valued/REAME.rst
@@ -1,0 +1,68 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Stock Picking Valued
+====================
+
+ * Add amount information to pickings. 
+ * Price and total per line from sale order.
+ * Total Amount of all lines.
+ * Select in partner and picking if you like show this information. 
+
+Installation
+============
+
+To install this module, you need to:
+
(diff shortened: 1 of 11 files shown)


clonedagain:
Technically this changes a "report" indeed, but it's not really reporting is it?
Maybe it would be best suited in stock-logistics-warehouse?

acysos:
Hello,

Yes, this module modify the stock.picking report, add prices, taxes and total to this report, if the partner has the option stock valued checked.

If you think that is better in other repository, no problem I close this PR and send a new to the other repository.

Greetings

pedrobaeza:
Well, I see that this repo is the more suitable. Equal to stock picking webkit report, that modifies standard one to make it work with Webkit engine.

Ignacio, I have to pass you one modified version of this module (including name), that enhances this one.

angelmoya:
I think will be better to do two modules:
- stock_valued, in stock-logistics-warehouse
- stock_valued_report, that depends to stock_valued, in this repo

pedrobaeza:
We have semi-migrated a better version of this module in my repo: https://github.com/serviciosbaeza/odoo-addons/pull/4. Please check that one, and let's complete it with the qweb report instead. This don't need the split, because you can choose by partner if you want a valued picking or not.

acysos:
I think that it's not neccesary split in two module. This module work and use the new api, but for the moment I don't continue the development, because I have other priority modules. If some one have time to merge the improvement that  say Pedro, it can be done.

gurneyalex:
:+1:

pedrobaeza:
As it seems there's no interest, I close. Feel free to reopen in case there's interest.

PR #17, opened 2015-04-01
7.0 cubep bill of lading webkit

mcassuto:


gurneyalex:
@mcassuto : your CLA is OK, the issue is with the two first commits which have no github login information.

PR #36, opened 2017-08-31
[FIX] Upgrade .yaml for correct CIs

chienandalu:
Upgrade .yaml so Runbot can build in PRs

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking this.

PR #41, opened 2017-12-16
[9.0][FIX] stock_valued_picking_report: Add ACL's for portal group

sergio-teruel:
Needed for portal users, without ACL they can't view computed values when the report is downloaded in website portal account.
cc @Tecnativa


pedrobaeza (approved):
Fast-tracking as obvious. Please do the same with 10.0

PR #48, opened 2018-11-21
[11.0][ADD][stock_product_last_usage_report]

JordiBForgeFlow:
This module adds the report *Latest Inventories & Moves* that was available
in previous versions of Odoo.

This report provides the ability to display, for each product, the latest
date it has been moved, or been involved in an inventory count.

It is useful to identify products that are no longer in use.

Usage
=====

Go to *Inventory / Reporting / Latest Inventories & Moves*


MiquelRForgeFlow:
all CIs are red :S

stock_product_last_usage_report/report/report_stock_lines_date.py
```diff
@@ -0,0 +1,65 @@
...
+from odoo import api, fields, models
+from odoo import tools
+
+
+class ReportStockLinesDate(models.Model):
+    _name = "report.stock.lines.date"
+    _description = "Dates of Inventories and latest Moves"
+    _auto = False
+    _order = "date"
+    id = fields.Integer('Product Id', readonly=True)
```
MiquelRForgeFlow, line 14: could you rename this field to another name different from 'id'?

stock_product_last_usage_report/report/report_stock_lines_date.py
```diff
@@ -0,0 +1,65 @@
...
+                                 index=True,
+                                 )
+    date = fields.Datetime('Date of latest Inventory', readonly=True)
+    move_date = fields.Datetime('Date of latest Stock Move', readonly=True)
+    active = fields.Boolean("Active", readonly=True)
+
+    @api.model_cr
+    def init(self):
+        tools.drop_view_if_exists(self.env.cr, self._table)
+        self.env.cr.execute("""
```
LoisRForgeFlow, line 25: You should put `# pylint: disable = E8103` before to avoid travis error.

AdriaGForgeFlow (commented):
IMO you should add an access rule as suggested by runbot.
E.g:
access_report_stock_lines_date,access_report_stock_lines_date,model_report_stock_lines_date,base.group_user,1,0,0,0

stock_product_last_usage_report/report/report_stock_lines_date.py
rousseldenis: @lreficent No! The error means that you have to pass parameters as real arguments to execute() function. @jbeficent Look for code, there are plenty of examples.

LoisRForgeFlow: here it was passed with args and yet the pyling disable was needed: https://github.com/OCA/stock-logistics-warehouse/blob/10.0/stock_inventory_exclude_sublocation/models/stock_inventory.py#L55

LoisRForgeFlow: plus here what is being inyected is this: `self._table,`. Do you prefer to hardcode the table name above as odoo does (https://github.com/odoo/odoo/blob/9.0/addons/stock/report/report_stock.py#L23)?

rousseldenis: > here it was passed with args and yet the pyling disable was needed: https://github.com/OCA/stock-logistics-warehouse/blob/10.0/stock_inventory_exclude_sublocation/models/stock_inventory.py#L55

This one is a bad example as domain is a string 'injected' in query

rousseldenis: No. Just don't pass arguments as string injection

LoisRForgeFlow: > This one is a bad example as domain is a string 'injected' in query

I see, that's right. Then we can go with the args in the execute function :+1:

rousseldenis:
@jbeficent Any news on this ?

PR #56, opened 2019-05-10
[FIX] stock_picking_report_valued: Wrong taxes amount if round globally

carlosdauden:
@Tecnativa

### DIFF
--- a/stock_picking_report_valued/models/stock_picking.py
+++ b/stock_picking_report_valued/models/stock_picking.py
@@ -41,12 +41,30 @@ class StockPicking(models.Model):
         records...).
         """
         for pick in self:
-            amount_untaxed = sum(pick.move_line_ids.mapped(
-                'sale_price_subtotal'))
-            amount_tax = sum(pick.move_line_ids.mapped(
-                'sale_price_tax'))
+            round_curr = self.sale_id.currency_id.round
+            amount_untaxed = amount_tax = 0.0
+            for tax_id, tax_group in self.get_taxes_values().items():
+                amount_untaxed += round_curr(tax_group['base'])
+                amount_tax += round_curr(tax_group['amount'])
             pick.update({
                 'amount_untaxed': amount_untaxed,
                 'amount_tax': amount_tax,
                 'amount_total': amount_untaxed + amount_tax,
             })
+
+    @api.multi
+    def get_taxes_values(self):
+        tax_grouped = {}
+        for line in self.move_line_ids:
+            tax = line.sale_line.tax_id
+            tax_id = tax.id
+            if tax_id not in tax_grouped:
+                tax_grouped[tax_id] = {
+                    'amount': line.sale_price_tax,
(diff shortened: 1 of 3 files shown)


stock_picking_report_valued/models/stock_picking.py
```diff
@@ -41,12 +41,30 @@ def _compute_amount_all(self):
         records...).
         """
         for pick in self:
-            amount_untaxed = sum(pick.move_line_ids.mapped(
-                'sale_price_subtotal'))
-            amount_tax = sum(pick.move_line_ids.mapped(
-                'sale_price_tax'))
+            round_curr = self.sale_id.currency_id.round
+            amount_untaxed = amount_tax = 0.0
+            for tax_id, tax_group in self.get_taxes_values().items():
```
rousseldenis: Maybe this computation on rounding should depends on config parameter (round per line or round globally) as the result should be different.

stock_picking_report_valued/models/stock_picking.py
```diff
@@ -41,12 +41,30 @@ def _compute_amount_all(self):
         records...).
         """
         for pick in self:
-            amount_untaxed = sum(pick.move_line_ids.mapped(
-                'sale_price_subtotal'))
-            amount_tax = sum(pick.move_line_ids.mapped(
-                'sale_price_tax'))
+            round_curr = self.sale_id.currency_id.round
```
rousseldenis:
```suggestion
            round_curr = pick.sale_id.currency_id.round
```
and so on.

rousseldenis (changes requested):

stock_picking_report_valued/models/stock_picking.py
carlosdauden: Changes done

stock_picking_report_valued/models/stock_picking.py
carlosdauden: Ups...
Changes done

rousseldenis: > and so on.

Missed the next self

carlosdauden: You are right, changed.
Thanks.

pedrobaeza (commented):
Travis is red. Seeing this, why don't we ignore stock.move.line fields and directly compute things at stock.move level and only checking qty at sml level?

pedrobaeza:
@carlosdauden what do you think about my thoughts?

carlosdauden:
@pedrobaeza 
As what is printed on the delivery note are stock.move.line if we computed the fields in stock.move then we would have to recalculate the proportional part of each printed line.
Another thing is that the Odoo core report printed the stock.move and added the lot inside `<td>`, which is how I think it should be...

pedrobaeza (approved):
OK, got it

rousseldenis (approved):
Code review

PR #59, opened 2019-06-11
[FIX] stock_picking_report_valued: Remove round_method condition because don't apply here

carlosdauden:
Never we can sum decimals of 2 distinct taxes and round after

round_method == 'round_globally'
base: 11.10   tax 4%: 0.444
base: 36.14  tax 10%: 3.614

Before this PR
Picking report totals > base: 47.24  taxes: 4.06  total: 51.30

After this PR
Picking report totals > base: 47.24  taxes: 4.05  total: 51.29

When invoice is created it shows always:
IVA 10% en 36,14 € 3,61 €
IVA 4% en 11,10 € 0,44 €

@Tecnativa

### DIFF
--- a/stock_picking_report_valued/__manifest__.py
+++ b/stock_picking_report_valued/__manifest__.py
@@ -8,7 +8,7 @@
 {
     "name": "Valued Picking Report",
     "summary": "Adding Valued Picking on Delivery Slip report",
-    "version": "11.0.1.0.1",
+    "version": "11.0.1.0.2",
     "author": "Tecnativa, "
               "Odoo Community Association (OCA)",
     "website": "https://www.tecnativa.com",
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -64,7 +64,8 @@ class StockMoveLine(models.Model):
         """
         for line in self:
             taxes = line.sale_tax_id.compute_all(
-                price_unit=line.sale_line.price_reduce,
+                price_unit=line.sale_line.price_subtotal / (
+                    line.sale_line.product_uom_qty or 0.0),
                 currency=line.currency_id,
                 quantity=line.qty_done or line.product_qty,
                 product=line.product_id,
--- a/stock_picking_report_valued/models/stock_picking.py
+++ b/stock_picking_report_valued/models/stock_picking.py
@@ -41,19 +41,11 @@ class StockPicking(models.Model):
         records...).
         """
         for pick in self:
-            sale = pick.sale_id
-            round_method = sale.company_id.tax_calculation_rounding_method
-            if round_method == 'round_globally':
-                amount_untaxed = sum(pick.move_line_ids.mapped(
-                    'sale_price_subtotal'))
-                amount_tax = sum(pick.move_line_ids.mapped(
-                    'sale_price_tax'))
-            else:
-                round_curr = sale.currency_id.round
-                amount_untaxed = amount_tax = 0.0
-                for tax_id, tax_group in pick.get_taxes_values().items():
-                    amount_untaxed += round_curr(tax_group['base'])
-                    amount_tax += round_curr(tax_group['amount'])
+            round_curr = pick.sale_id.currency_id.round
+            amount_untaxed = amount_tax = 0.0
+            for tax_id, tax_group in pick.get_taxes_values().items():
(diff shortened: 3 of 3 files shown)


rousseldenis:
As tax computation on sale orders depends on accounting tax computation, your assertion is false.

I've tested on runbot odoo v11 with both parameters.

The result is as expected.

With round_globally :

image

With round per line:

image


So, the report should be printed accordingly

pedrobaeza:
@rousseldenis but have you tried on pickings values (which is the aim of this module)? Your screenshots are from sales orders.

rousseldenis:
> @rousseldenis but have you tried on pickings values (which is the aim of this module)? Your screenshots are from sales orders.

Ok, didn't see it was computed from move line value (which has already done the separate computation).

pedrobaeza (approved):

rousseldenis:
Can you check almost if both flows are correctly computed (sale order > picking > picking report) ?

pedrobaeza:
> Can you check almost if both flows are correctly computed (sale order > picking > picking report) ?

Are you telling me? This has been tried in customer's DB

rousseldenis:
> Are you telling me? This has been tried in customer's DB

Yes (or @carlosdauden which is the same) :smiley: 
With both tax computations ? (per line and globally) ?

pedrobaeza:
We have tried one test case just now, but it was the one who was failing before. @carlosdauden can you confirm the other method doesn't have any problem?

carlosdauden:
The real problem is that the order itself is poorly calculated.
But that must be another PR to odoo.

@rousseldenis Can you try to invoice the attach sale order with round_globally? :wink: 


Selección_227
Selección_228

rousseldenis:
> The real problem is that the order itself is poorly calculated.
> But that must be another PR to odoo.

If you set a different method between your two screenshots, that's a **normal** behaviour and has not to be changed. This is not a bug.

pedrobaeza:
@rousseldenis the problem is that Odoo itself put a different value in sales order and in invoice, and this change what makes is to mimic the invoice behavior, not the sales order one.

carlosdauden:
The problem is not based so much on imitating one behavior or another, but what the sale order total taxes compute is wrong.
As I said before, we can not add different tax decimals residual.
The grouped total of the taxes should always be rounded to the decimals of the currency.

We can have 50 lines with a 4% tax and another 30 with a 10% tax. and if the sum of these quotas (once one type of rounding or another is applied) results in
10,273 for 4% and 30,003 for 10% the total taxes can not be 40.28 in any case.
Because the taxes are 10.27 + 30.00 = 40.27

Sorry, I do not know if I've explained myself better now.

emagdalenaC2i:
Tested in runbot. Seems that there are still some issues.
I am using discounts and Round Globally method
image

image

pedrobaeza:
@emagdalenaC2i can you check the created invoice?

emagdalenaC2i:
Here it is 
image

pedrobaeza:
OK, then the picking is indeed with other value. I'll check this with @carlosdauden tomorrow to fully understand the problem.

rousseldenis:
There was a discussion on a PR https://github.com/odoo/odoo/pull/28788

But I didn't investiguate. The aim is to have the same behaviour than Odoo

carlosdauden:
@emagdalenaC2i Last commit resolves "unrelated" problem

price_subtotal field is computed with price with complete decimal precision:
https://github.com/odoo/odoo/blob/a25f6e1b9617a04ccc517ed2e53429ca233ba728/addons/sale/models/sale.py#L727

price_reduce is computed with complete decimal precision but when it is stored it is rounded to product decimal precision
https://github.com/odoo/odoo/blob/a25f6e1b9617a04ccc517ed2e53429ca233ba728/addons/sale/models/sale.py#L824
https://github.com/odoo/odoo/blob/a25f6e1b9617a04ccc517ed2e53429ca233ba728/addons/sale/models/sale.py#L927

emagdalenaC2i (approved):
Review and tested in Runbot

emagdalenaC2i:
@pedrobaeza I think this is ready now

pedrobaeza:
Clarified with @carlosdauden. This patch is correct, and we will raise 2 issues to Odoo about incoherences with:

- price_reduce precision
- tax totals in sales orders.

Merging this.

PR #70, opened 2019-08-08
[IMP] stock_picking_report_valued: Value fields showed only in customer pickings

carlosdauden:
@Tecnativa


pedrobaeza:
What happened with dropshipping pickings?

If I understand well, the condition of checking `sale_id` already restricts the view properly. I think the problem is in the header column, that is not hidden. Can it be?

PR #76, opened 2019-10-28
[ADD] stock_move_value_report: New module

chienandalu:
This module allows to obtain a report with the cost value of a given
operation set like a picking, a scrap, an inventory or even a selection of
moves or move lines.

cc @Tecnativa TT18619

### DIFF
--- a/stock_move_value_report/__manifest__.py
+++ b/stock_move_value_report/__manifest__.py
@@ -1,0 +1,24 @@
+# Copyright 2019 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock Move Cost Value Report",
+    "version": "11.0.1.0.0",
+    "category": "Warehouse",
+    "website": "https://github.com/OCA/stock-logistics-reporting",
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "stock_account",
+    ],
+    "data": [
+        'report/stock_value_report.xml',
+        'report/stock_move_line_value_report.xml',
+        'report/stock_move_value_report.xml',
+        'report/stock_picking_value_report.xml',
+        'report/stock_scrap_value_report.xml',
+        'report/stock_inventory_value_report.xml',
+    ],
+}
--- a/stock_move_value_report/report/stock_move_line_value_report.xml
+++ b/stock_move_value_report/report/stock_move_line_value_report.xml
@@ -1,0 +1,166 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <template id="stock_move_lines_value">
+        <table class="table table-condensed mt4 mb0">
+            <t t-set="has_serial_number" t-value="move_lines.mapped('lot_id')" groups="stock.group_production_lot"/>
+            <thead>
+                <tr>
+                    <th name="th_product">
+                        <strong>Product</strong>
+                    </th>
+                    <th name="th_lot" t-if="has_serial_number">
+                        Lot/Serial Number
+                    </th>
+                    <th name="th_qty" class="text-center">
+                        <strong>Quantity</strong>
+                    </th>
+                    <th name="th_value" class="text-right">
+                        <strong>Value</strong>
+                    </th>
+                    <th name="th_value_sbtl" class="text-right">
+                        <strong>Subtotal</strong>
+                    </th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr t-foreach="move_lines" t-as="move_line">
+                    <td>
+                        <span t-field="move_line.product_id"/>
+                        <span t-field="move_line.product_id.sudo().description"/>
+                    </td>
+                    <t t-if="has_serial_number">
+                        <td>
+                            <table width="100%">
+                                <tr>
+                                    <td>
+                                        <span t-field="move_line.lot_id"/>
+                                        <t t-if="not move_line.lot_id">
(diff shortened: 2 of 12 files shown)


i18n/es.po
```diff
@@ -0,0 +1,172 @@
+# Translation of Odoo Server.
```
pedrobaeza: Incorrect path

chienandalu: Oops. Fixed :)

stock_move_value_report/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2019 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock Move Cost Value Report",
+    "version": "11.0.1.0.0",
+    "category": "Stock",
```
rousseldenis: Set it to 'Warehouse' as it won't appear in good category on OCA shop.

stock_move_value_report/report/stock_move_line_value_report.xml
```diff
@@ -0,0 +1,166 @@
...
+                        Lot/Serial Number
+                    </th>
+                    <th name="th_qty" class="text-center">
+                        <strong>Quantity</strong>
+                    </th>
+                    <th name="th_value" class="text-right">
+                        <strong>Value</strong>
+                    </th>
+                    <th name="th_value_sbtl" class="text-right">
+                        <strong>Value Subtotal</strong>
```
rousseldenis: IMHO 'Subtotal' should be sufficeint

stock_move_value_report/report/stock_move_line_value_report.xml
```diff
@@ -0,0 +1,166 @@
...
+                        </t>
+                    </t>
+                </div>
+                <!-- Total all moves -->
+                <div class="clearfix">
+                    <div id="total" class="row">
+                        <div class="col-xs-4 pull-right">
+                            <table class="table table-condensed" style="min-width: 250px;max-width: 400px;page-break-inside: avoid;">
+                                <tr class="border-black">
+                                    <td><h4>Total operations value balance</h4></td>
```
rousseldenis: This one is too big related to the 'Total' line.

image

stock_move_value_report/report/stock_move_value_report.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <template id="report_stock_move_value">
+        <t t-call="stock_move_value_report.report_stock_move_value_document">
+            <t t-set="move_lines_ids" t-value="docs.mapped('move_line_ids')"/>
```
rousseldenis, line 6: This works only if 'Show detailed operations' is checked on picking type. Either the report is void.

stock_move_value_report/report/stock_picking_value_report.xml
```diff
@@ -0,0 +1,88 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <template id="report_stock_picking_value_document">
+        <t t-call="web.html_container">
+            <t t-call="web.external_layout">
+                <div class="page">
+                    <div class="row">
+                        <div class="col-xs-6">
+                            <div t-if="picking.move_lines and picking.move_lines[0].partner_id and picking.move_lines[0].partner_id.id != picking.partner_id.id">
```
rousseldenis, line 10: As above, this works only if you checked the detailed operations.

rousseldenis (changes requested):
Code and functional review.

Missing the case if move_lines are not used by config.

stock_move_value_report/report/stock_move_value_report.xml
chienandalu: No, it isn't. The stock.move.line is created as soon as move is reserved or done even if you don't see it. Indeed what you see in the stock picking delivery slip is the move lines created.

stock_move_value_report/report/stock_picking_value_report.xml
chienandalu: Same here.

stock_move_value_report/report/stock_move_line_value_report.xml
chienandalu: Well, in cases like AVCO, the customer wants to see the price unit of the operation so it's clear at what cost the operation was made

stock_move_value_report/report/stock_move_value_report.xml
rousseldenis: Of course, my bad. Didn't check the state of picking.

stock_move_value_report/report/stock_move_line_value_report.xml
chienandalu: Well, it's to make a distinction with the partials coming from pickings, inventories or scraps from the move lines selected

stock_move_value_report/report/stock_move_line_value_report.xml
rousseldenis: It was just for the string title. On the report, it's quite big.

stock_move_value_report/report/stock_move_line_value_report.xml
rousseldenis: Yes, but it's a little bit ugly, no ?

rafaelbn (approved):
Tested in runbot :+1: 

@rousseldenis is it ok for you?

stock_move_value_report/report/stock_move_line_value_report.xml
chienandalu: Ok, I changed the layout a bit to fit in a more integrated way.

rousseldenis (approved):
Great!

PR #78, opened 2019-11-12
[12.0][ADD] stock_picking_product_label

NuriaXifre:
New module that allows printing labels of all products in a stock picking. 
cc @eficent @joansis


pedrobaeza:
Please avoid plurals in module name.

PR #80, opened 2019-11-14
[FIX] stock_quantity_history_location: Wrong Dependency in stock_quantity_history_location

ougc27:
Hello

This Pull request fix a bug on the valuation report  using the module stock_quantity_history_location.

Before  2f44e53 the direct module dependency was stock, this provoked that super call of open_table method of the wizard never executed.

I appreciate your support in advance.


rafaelbn (approved):
Functionally tested :+1:

FIX the issue functionally.

pedrobaeza:
It's supposed that the glue module https://github.com/OCA/stock-logistics-reporting/tree/12.0/stock_account_quantity_history_location is for this.

PR #86, opened 2019-12-30
[FIX] stock_quantity_history_location: qty by location

chienandalu:
When a qty at date is asked the location context could be overriden if
the company_owned one was inserted after we removed it. We ensure to
remove it to always obtain the right values.

As a side effect, stock_account valuation is now correctly computed when
AVCO or Standard methods are implied. So should be with FIFO manual (as
it computes the quantities the same way) but doubtful in FIFO real time.

So maybe stock_account_quantity_history_location is of no much use (warning advising that the computation is not true, wich isn't accurate and show 0 qtys values to dectect errors)

cc @Tecnativa TT21183

### DIFF
--- a/stock_quantity_history_location/models/product.py
+++ b/stock_quantity_history_location/models/product.py
@@ -1,0 +1,16 @@
+# Copyright 2019 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def _get_domain_locations(self):
+        """Ensures the location context isn't overriden"""
+        ctx = dict(self.env.context)
+        if not ctx.get('drop_company_owned') or not ctx.get('company_owned'):
+            return super()._get_domain_locations()
+        ctx.pop('company_owned', None)
+        self_ctx = self.with_context(ctx)
+        return super(ProductProduct, self_ctx)._get_domain_locations()
--- a/stock_account_quantity_history_location/tests/__init__.py
+++ b/stock_account_quantity_history_location/tests/__init__.py
@@ -1,0 +1,1 @@
+from . import test_valuation_location
--- a/stock_account_quantity_history_location/tests/test_valuation_location.py
+++ b/stock_account_quantity_history_location/tests/test_valuation_location.py
@@ -1,0 +1,55 @@
+# Copyright 2020 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo.addons.stock_quantity_history_location.tests import (
+    test_stock_quantity_history_location
+)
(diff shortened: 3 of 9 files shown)


chienandalu:
@rafaelbn please test

stock_account_quantity_history_location/__manifest__.py
```diff
@@ -19,6 +19,6 @@
     'data': [
         'wizards/stock_quantity_history.xml',
     ],
-    'auto_install': True,
+    'auto_install': False,
```
pedrobaeza: This is correct to be auto-installed when both dependencies are there

chienandalu: ok, I'll keep it like it was. Only this module maybe isn't pertinent anymore, as it doesn't solve anything that cannot provide this one...

pedrobaeza: Then remove it. cc @JordiBForgeFlow

chienandalu: Sorry, I was wrong. There's at least a view that need to be covered. So I just keep it as it was.

stock_quantity_history_location/models/product.py
```diff
@@ -0,0 +1,16 @@
...
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def _get_domain_locations(self):
+        """Ensures the location context isn't overriden"""
+        ctx = dict(self.env.context)
+        if not ctx.get('drop_company_owned') or not ctx.get('company_owned'):
+            return super()._get_domain_locations()
+        ctx.pop('company_owned')
```
yajo:
```suggestion
        ctx.pop('company_owned', None)
```

rafaelbn (approved):
Functionally tested :+1:  :smile:

stock_quantity_history_location/models/product.py
chienandalu: Mmm, what for? I just want to drop it so there's no need of any default value

chienandalu: Mmm, what for? I really don't care about the value

yajo: Sorry, to avoid a `KeyError` if it's missing.

rousseldenis:
@chienandalu Do you plan to add a test for that fix ?

chienandalu:
@rousseldenis Yes, I'd like to

dreispt (approved):

stock_quantity_history_location/models/product.py
chienandalu: Done :)

chienandalu:
@rousseldenis Added tests

rousseldenis (approved):
Code review

rousseldenis:
@chienandalu Is this ready ?

chienandalu:
Yeah :)

PR #89, opened 2020-02-02
[FIX] Standard price incorrectly calculated on historic value

CasVissers-360ERP:
When creating an historic inventory value export to XLS the current cost price of the product was used. This fix proposes using the historic cost price and changing the calculation of the value.

@ps-tubtim @rafaelbn


dreispt (approved):

ps-tubtim (approved):
Code LGTM

PR #93, opened 2020-02-26
12.0 stock kardex

Tino480:


stock_kardex_report/models/stock_kardex_report.py
```diff
@@ -0,0 +1,163 @@
+# Copyright 2020, Jarsa Sistemas, S.A. de C.V.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import _, fields, models
+import textwrap
+
+
+class StockKardexReportWiz(models.TransientModel):
```
alan196: If it's a wizard it must be in a separate folder.. see [module structure](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#17complete-structure)

alan196 (changes requested):
@Tino480 Please see Travis error, you need to add a security rule to the new model, define it to the Inventory / User group.

stock_kardex_report/models/stock_kardex_report.py
Tino480: Created folders

stock_kardex_report/wizard/stock_kardex_report_wizard.py
```diff
@@ -0,0 +1,144 @@
...
+            'lot_id': False,
+            'owner_id': False,
+            'package_id': False,
+            'qty_done': 0,
+            'location_id': False,
+            'location_dest_id': False,
+            'date': self.date_from,
+            'origin': 'Initial Balance',
+            'balance': total,
+        })
```
alan196:
```suggestion
        report_list.append({
            'product_id': self.product.id,
            'product_uom_id': product_id.uom_id.id,
            'date': self.date_from,
            'origin': _('Initial Balance'),
            'balance': total,
        })
```

If you don't use the fields you don't need to define it as False

alan196 (changes requested):

dreispt:
Con you provide some context? What is "kardex"?

alan196:
@dreispt 
I really don't know the term in English, I have heard about it only in Mexico.

I will explain what this report shows and maybe you can help us to name it properly.

This report shows the stock of a product in a specific location in a defined time period.

If I ask a report for Product A from 01-01-2020 to 03-01-2020 in location WH/Stock

The report will be the following

| Product | Date |  Description | Origin | Destination | Qty | Balance |
| --- | --- | --- | --- | --- |  --- | --- |
| Product A | 01-01-2020 12:00:00 | Initial Balance |  |  | 100 | 100 |
| Product A | 02-01-2020 12:00:00 | PO001 | Vendors | WH/Stock | 100 | 200 |
| Product A | 02-01-2020 14:00:00 | SO001 | WH/Stock | Customers | -50 | 150 |
| Product A | 02-01-2020 16:30:00 | SO002 | WH/Stock | Customers | -75 | 75 |
| Product A | 03-01-2020 12:00:00 | PO002 | Vendors | WH/Stock | 100 | 175 |

dreispt:
Thank for the clarification.
Do you mind copying that explanation you just have into the README Description?

alan196:
@dreispt done

stock_kardex_report/wizard/stock_kardex_report_wizard.py
```diff
@@ -0,0 +1,139 @@
...
+    date_from = fields.Datetime(
+        string='From', required=True, default=fields.Datetime.now)
+    date_to = fields.Datetime(
+        string='To', required=True, default=fields.Datetime.now)
+    product = fields.Many2one('product.product', required=True)
+    location = fields.Many2one('stock.location', required=True)
+
+    def open_table(self):
+        self.env['stock.kardex.report'].search([]).unlink()
+        self._cr.execute('''
```
dreispt, line 21: You could have used double quotes here, to then cleanly use single quotes in the SQL.

dreispt (approved):
Code review, LGTM

alan196 (approved):
Functional and technical review

rousseldenis (approved):

rousseldenis:
@Tino480 Have you see that this module exist in v13 ? https://github.com/OCA/stock-logistics-reporting/tree/13.0/stock_card_report

rousseldenis:
And in v12, in fact... So, I think we cannot merge this as it is a duplicate in a funcitonal point of view. @dreispt

dreispt:
Thank you @rousseldenis 
@Tino480 in general we prefer to not have modules with duplicate features. Can you check the the module pointed out by @rousseldenis ?

mileo:
Hi, Nice job! but i thin i found a issue in the initial balance:

image

image

image

Another thing It would not be better to extend it in another module or even improve the existing report of the stock_card_report 

image

image

mileo (changes requested):
Please check the comment of the initial balance

PR #95, opened 2020-03-05
[12.0][IMP] stock_inventory_valuation_report: Reference and Barcode columns

ivantodorovich:
It separates the reference and name into 2 columns (instead of showing the display name).
This is usually better for analysis, specially on the excel file.

It also adds the barcode to the report, which is usefull most of the times

### DIFF
--- a/stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
+++ b/stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
@@ -8,7 +8,9 @@ class StockInventoryValuationView(models.TransientModel):
     _name = 'stock.inventory.valuation.view'
     _description = 'Stock Inventory Valuation View'
 
-    display_name = fields.Char()
+    name = fields.Char()
+    reference = fields.Char()
+    barcode = fields.Char()
     qty_at_date = fields.Float()
     uom_id = fields.Many2one(
         comodel_name='uom.uom',
@@ -54,7 +56,9 @@ class StockInventoryValuationReport(models.TransientModel):
         ReportLine = self.env['stock.inventory.valuation.view']
         for product in products:
             line = {
-                'display_name': product.display_name,
+                'name': product.name,
+                'reference': product.default_code,
+                'barcode': product.barcode,
                 'qty_at_date': product.qty_at_date,
                 'uom_id': product.uom_id,
                 'currency_id': product.currency_id,
--- a/stock_inventory_valuation_report/reports/stock_inventory_valuation_report.xml
+++ b/stock_inventory_valuation_report/reports/stock_inventory_valuation_report.xml
@@ -99,7 +99,9 @@
         <div class="act_as_thead">
             <div class="act_as_row labels">
                 <div class="act_as_cell">#</div>
-                <div class="act_as_cell">Display Name</div>
+                <div class="act_as_cell">Reference</div>
+                <div class="act_as_cell">Name</div>
+                <div class="act_as_cell">Barcode</div>
                 <div class="act_as_cell">Quantity</div>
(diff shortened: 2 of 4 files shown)


ivantodorovich:
ping @ps-tubtim

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
```diff
@@ -9,6 +9,9 @@ class StockInventoryValuationView(models.TransientModel):
     _description = 'Stock Inventory Valuation View'
 
     display_name = fields.Char()
```
ps-tubtim:
```suggestion

```
If this model has `name` and `reference`, `display_name` is unnecessary.

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
```diff
@@ -55,6 +58,9 @@ def _compute_results(self):
         for product in products:
             line = {
                 'display_name': product.display_name,
```
ps-tubtim:
```suggestion

```
Like above comment.

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.xml
```diff
@@ -113,7 +115,13 @@
...
             <div class="act_as_cell left">
-                <t t-esc="line.display_name"/>
+                <t t-esc="line.reference"/>
+            </div>
+            <div class="act_as_cell left">
+                <t t-esc="line.name"/>
+            </div>
+            <div class="act_as_cell left">
+                <t t-esc="line.barcode"/>
             </div>
```
ps-tubtim, line 125: I think this report should show barcode image. Could you please do this?

ps-tubtim (commented):
@ivantodorovich I think this is good ideas, but I have some comments.

ps-tubtim:
Functional Test: Total of Value in xlsx report show wrong column.

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.xml
ivantodorovich: I tried but it doesn't look good. Either we use big rows, or the barcode is too small.
Anyways, who will use it to scan? It's an inventory valuation, barcode is usually used for reference in that case.

Our customer's only use case is to show the barcode numbers, not the picture

image

ivantodorovich:
> Functional Test: Total of Value in xlsx report show wrong column.

Resolved in new commit

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
ivantodorovich: I would keep it in case someone wants to extend the view and use it.
If it's blocking I'll remove

ps-tubtim: IMO, `name` and `reference` can use instead of `display_name`.

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.xml
ps-tubtim: Ok, I agree on it. So this report could show the barcode numbers.

ps-tubtim:
@ivantodorovich Please squash commits into one.

ivantodorovich:
> @ivantodorovich Please squash commits into one.

done :)

ps-tubtim (approved):
:+1:

dreispt (approved):

dreispt:
Hello, #99 needs some love!

PR #99, opened 2020-03-27
[12.0][ADD] Inventory Turnover Report

dreispt:


stock_inventory_turnover_report/models/product_product.py
```diff
@@ -0,0 +1,122 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from dateutil.relativedelta import relativedelta
+from odoo import fields, models
+
+
+class Product(models.Model):
+    _inherit = "product.product"
+
+    def _compute_quantities_in_dict(self, from_date):
```
ivantodorovich, line 11: qty_available already considers locations in the context (gotten from the search view). This allows to analyze data by location, warehouse, company, etc..

 https://github.com/odoo/odoo/blob/7647077e9762359060b408fa923773826998fedf/addons/stock/models/product.py#L153

It'd be good to implement the same in this method. If you do this then this report could be filtered by location or warehouse :thinking: otherwise it's probably giving wrong data if any of these filters are active

ivantodorovich (changes requested):

smangukiya (approved):
Code looks good to me, though I am agreed with @ivantodorovich 's comment where if we can generalize the filter domain based on the warehouse, location, etc then it would be more useful.

dreispt:
@ivantodorovich @smangukiya I added that, although I'm not sure it is really usable right now.
How would the user take advantage of these context filters?

ivantodorovich:
> @ivantodorovich @smangukiya I added that, although I'm not sure it is really usable right now.
> How would the user take advantage of these context filters?

By selecting the locations / warehouse on the search view :)
image

```xml
<?xml version="1.0"?>
<xpath expr="//field[@name='pricelist_id']" position="after">
                    <separator/>
                    <field name="location_id" widget="selection" context="{'location': self}" filter_domain="[]"/>
                    <field name="warehouse_id" widget="selection" context="{'warehouse': self}" filter_domain="[]"/>
                    <separator/>
                    <filter name="real_stock_available" string="Available Products" domain="[('qty_available','&gt;',0)]"/>
                    <filter name="real_stock_negative" string="Negative Forecasted Quantity" domain="[('virtual_available','&lt;',0)]"/>
                </xpath>
            
```

ivantodorovich (approved):

bodedra (approved):

PR #120, opened 2020-10-05
[13.0][FIX] stock_picking_report_valued: Make visible valued_picking field in customer

sergio-teruel:
CC @Tecnativa TT25999
Since customer field disappear I think that valued_picking field must be always visible.


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this as obvious patch:

/ocabot merge minor

PR #125, opened 2020-11-23
[ADD] stock_picking_report_valued_sale_mrp: New module

chienandalu:
image

cc @Tecnativa TT26711


stock_picking_report_valued_sale_mrp/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "version": "12.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/stock-logistics-reporting",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "maintainers": ["chienandalu"],
+    "license": "AGPL-3",
+    "depends": [
+        "stock_picking_report_valued",
+        "mrp",
```
pedrobaeza:
```suggestion
        "sale_mrp",
```

stock_picking_report_valued_sale_mrp/models/stock_move_line.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class StockMoveLine(models.Model):
+    _inherit = 'stock.move.line'
+
+    phantom_product_id = fields.Many2one(
+        comodel_name="product.product",
+        compute="_compute_phantom_product_id",
+        compute_sudo=True,
+        string="Kit product",
```
pedrobaeza:
```suggestion
        string="Product Kit",
```

stock_picking_report_valued_sale_mrp/models/stock_move_line.py
```diff
@@ -0,0 +1,90 @@
...
+        readonly=True,
+    )
+    phantom_line = fields.Boolean(
+        compute="_compute_sale_order_line_fields",
+        compute_sudo=True,
+        default=False,
```
pedrobaeza: Why this default?

stock_picking_report_valued_sale_mrp/models/stock_move_line.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    phantom_dlvd_qty = fields.Float(
```
pedrobaeza: Don't use such cryptic abbreviates:
```suggestion
    phantom_delivered_qty = fields.Float(
```

stock_picking_report_valued_sale_mrp/models/stock_move_line.py
```diff
@@ -0,0 +1,90 @@
...
+        kit_lines = self.filtered("phantom_product_id")
+        for sale_line in kit_lines.mapped("sale_line"):
+            move_lines = kit_lines.filtered(
+                lambda x: x.sale_line == sale_line)
+            phantom_line = move_lines[:1]
+            price_unit = (
+                sale_line.price_subtotal / sale_line.product_uom_qty
+                if sale_line.product_uom_qty else sale_line.price_reduce)
+            # Get the kit. We should store the one used somewhere
+            bom = self.env['mrp.bom'].sudo()._bom_find(
```
pedrobaeza: Isn't there any other mechanism? Why do we need to get the BoM? The idea is to get the information from the sale line. The problem if not is that there can be different BoMs when confirming the sales order than when printing the report.

pedrobaeza (changes requested):

stock_picking_report_valued_sale_mrp/models/stock_move_line.py
chienandalu: Yes, this is the main issue here. We don't have any info of the BoM used but we must figure out how many kits were delivered in the present picking.

If we assume the kits are inmutable (the components don't change in time), we could store them in the `sale.order.line` and then findout how many kits were delivered from the `stock.move.line`. Anyway, wouldn't that be a separate module?

pedrobaeza: I don't think we need such bom linking. If you have the linked sale.order.line, we have the original kit quantity, and dividing by the whole pickings quantity (for considering backorders), we have the factor.

chienandalu: Mmm, what happens if there are returns and redelivers?

pedrobaeza: You can control that through the many2one field that points to the original/returned picking.

chienandalu: I finding a big limitations with this aproach. Odoo can't tell very well how many kits were delivered when the picking is partial or a return is implied.

For example:

- I order 5 kits but I deliver just four and leave a backorder for the rest. Sale order line qty delivered = 0
- The I cancel the backorder: sale order line qty delivered = 5!

This limitation comes from the poor calculation that Odoo makes for these products:

https://github.com/odoo/odoo/blob/12.0/addons/sale_mrp/models/sale_mrp.py#L30-L33

Indeed, the aproach in v13 is precisely to link the boms with the moves: https://github.com/odoo/odoo/blob/13.0/addons/sale_mrp/models/sale_mrp.py#L25 and then compute according to the delivered component quantities.

chienandalu: Changes done

chienandalu:
Travis lint seems to fail for unrelated reasons

pedrobaeza:
Please rebase for having green linter.

ernestotejeda (approved):

carlosdauden (approved):

PR #128, opened 2020-11-30
[12.0][ADD] Report picking summary

quentinDupont:
This module adds the possibility to select several Pickings and print a Report summary with two sheets :
* Products Summary (where the total sum is calculated for each product)
* Detail Part (each Picking is shown)

pdf_report

Some options on wizard : 

image

Migration and improvment of @grap custom module from @legalsylvain : https://github.com/grap/grap-odoo-custom/tree/8.0/grap_qweb_report


quentinDupont:
Travis fails for other module : 
image

Issue #115, opened 2020-09-16
[12.0] [FIX] Tax Rounding in stock_picking_report_valued


emagdalenaC2i:
To replicate this issue:

1 - Set the Round globally option (Invoicing settings) and allow discounts (sale settings)

2- Create the next quotation
![image](https://user-images.githubusercontent.com/24428859/93372216-de134400-f853-11ea-9ce5-4b160baccb04.png)

3 - Confirm the quotation and check the differences in the quotation, invoice and the Delivery Slip report
![image](https://user-images.githubusercontent.com/24428859/93372434-2fbbce80-f854-11ea-906f-27c6ec04ab24.png)

pedrobaeza:
Odoo version? What is the tax total in invoices?

emagdalenaC2i:
I think this affect all versions but I check it with 12.0
Invoices taxes and total are the same as in the quotation

emagdalenaC2i:
This is the invoice 
![image](https://user-images.githubusercontent.com/24428859/93373299-73630800-f855-11ea-859a-5803ea552700.png)
You can check it in runbot https://runbot.odoo-community.org/runbot/151/12.0

emagdalenaC2i:
You should check https://github.com/OCA/account-invoicing/pull/704

emagdalenaC2i:
Here you can see other example:
To replicate this issue:

1 - Set the **Round per line** option (Invoicing settings) and allow discounts (sale settings)

2- Create the next quotation
![image](https://user-images.githubusercontent.com/24428859/93576861-a4e5eb80-f99b-11ea-97fe-5ed696898d3d.png)

3 - Confirm the quotation and check the differences in the quotation, invoice and the Delivery Slip report
![image](https://user-images.githubusercontent.com/24428859/93576958-c3e47d80-f99b-11ea-94db-94fa8ef4b65d.png)

Here is the invoice
![image](https://user-images.githubusercontent.com/24428859/93577729-b380d280-f99c-11ea-9025-9ba00af57204.png)

So, this is not a Round globally issue.

PR #130, opened 2021-01-03
[12.0][FIX] stock_picking_report_valued: Round computed price to avoid difference

carlosdauden:
between picking and invoice.

Steps to reproduce:
- Create a sale line with qty = 3.72 and price = 23.45 
- Confirm Sale Order and click in delivery
- Fill qty done = 3.55
- Print valued delivery

Before this PR: 
subtotal = 83.24 (3.55 * 23.448924731182796)

After this PR:
subtotal = 83.25 (3.55 * 23.45)

@Tecnativa TT17492

### DIFF
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -1,10 +1,11 @@
 # Copyright 2014-2018 Tecnativa - Pedro M. Baeza
 # Copyright 2015 Antonio Espinosa - Tecnativa <antonio.espinosa@tecnativa.com>
 # Copyright 2018 Luis M. Ontalba - Tecnativa <luis.martinez@tecnativa.com>
-# Copyright 2016-2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# Copyright 2016-2021 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can you please include a regression test that exercises the problem?

carlosdauden:
Test added

pedrobaeza (approved):

sergio-teruel (approved):

pedrobaeza:
My question in 13.0 is also relevant here: https://github.com/OCA/stock-logistics-reporting/pull/131#issuecomment-753804905

emagdalenaC2i:
This PR doesn't fix #115 and that should be fixed
image
image
image

emagdalenaC2i:
You can use your own example:
(if you set the tax computation as global)
image
image
image

emagdalenaC2i:
And with my other example:
(if you set the tax computation as per line) So, this is not a Round globally issue.
image
image
image

emagdalenaC2i:
@dreispt did you see my comments?

dreispt:
@emagdalenaC2i I0m sorry, I didn't understand there was a blocking issue.
AFAIU there is a open bug that it is yet not fixed in PR, not clear that that should block it,
 Please next time add a review or a "Needs Fixing" tag so that it is clear that the PR should not be merged.

PR #141, opened 2021-04-22
[14.0][IMP] stock_card_report, add  origin to display_name on report

kittiu:
Just add picking origin to the result

image

@ps-tubtim


ps-tubtim (approved):

newtratip (approved):
👌

PR #142, opened 2021-04-22
[FIX] stock_picking_report_valued_sale_mrp: ignored cancelled order

chienandalu:
We want to consider backordered moves even if they're cancelled but not
cancelled moves from cancelled orders, that must be filtered. Otherwise,
they pullute the amounts computation.

cc @Tecnativa TT29363

please review @ernestotejeda @pedrobaeza


pedrobaeza (approved):

chienandalu:
ping @carlosdauden

carlosdauden (approved):

PR #146, opened 2021-05-03
[13.0][MIG] stock_account_quantity_history_location: Migration to 13.0

victoralmau:
Migration to 13.0.

Locked by:
- [x] https://github.com/OCA/stock-logistics-reporting/pull/148

Please @chienandalu and @joao-p-marques can you review it?

@Tecnativa TT26940


stock_account_quantity_history_location/__manifest__.py
```diff
@@ -1,17 +1,18 @@
...
-    "version": "12.0.1.1.0",
+    "version": "13.0.1.0.0",
     "license": "AGPL-3",
-    "author": "Eficent," "Odoo Community Association (OCA)",
+    "author": "Eficent, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-reporting",
-    "depends": ["stock_account", "stock_quantity_history_location",],
-    "data": ["wizards/stock_quantity_history.xml",],
+    "depends": ["stock_account", "stock_quantity_history_location"],
+    "data": [],
```
joao-p-marques:
```suggestion
```

Could you explain why this wizard inheritance is no longer necessary?

joao-p-marques (commented):
Code review

stock_account_quantity_history_location/__manifest__.py
victoralmau: Now is not necessary to add wizard because `stock_quantity_history_location` addon add `location_id` in wizard, and i need to prevent duplication.

joao-p-marques: Ok :ok_hand: 

Can you attend my suggestion, then?

Also, you can squash the bot commits to reduce noise...

joao-p-marques (approved):

chienandalu (commented):
Isn't the point of this module to be able to filter the valuation (now called `stock.valuation.layer`) by location?
image

victoralmau:
> Isn't the point of this module to be able to filter the valuation (now called `stock.valuation.layer`) by location?
> image

Thanks, i have restored wizard view but IMO we could remove it if we remove these invisible attribute https://github.com/OCA/stock-logistics-reporting/blob/13.0/stock_quantity_history_location/wizards/stock_quantity_history.xml#L25
What do you think about it?

Other idea about `stock_quantity_history_location` addon, what do you think about possibility to include these functionality in `stock_quantity_history_location` addon?

chienandalu:
> Thanks, i have restored wizard view but IMO we could remove it if we remove these invisible attribute https://github.com/OCA/stock-logistics-reporting/blob/13.0/stock_quantity_history_location/wizards/stock_quantity_history.xml#L25
> What do you think about it?

Yes, well. It could be another approach, although:

> Other idea about `stock_quantity_history_location` addon, what do you think about possibility to include these functionality in `stock_quantity_history_location` addon?

The reason of this module is that tha stock valuation part is in the `stock_account` module, and that dependency isn't available in `stock_quantity_history_location`.

So may it's ok to have a proper view override.

Anyway, as long as I tested it (editing the view), I couldn't obtain the stock value by location, which should be the case. I'll do a further test in runbot once is built, although a test on the resultant `stock.valuation.layer` records values would be welcome

pedrobaeza:
@chienandalu can you re-review?

stock_account_quantity_history_location/wizards/stock_quantity_history.xml
```diff
@@ -0,0 +1,16 @@
...
+<odoo>
+    <record id="view_stock_account_quantity_history_location" model="ir.ui.view">
+        <field name="name">Valuation Report</field>
+        <field name="model">stock.quantity.history</field>
+        <field name="inherit_id" ref="stock.view_stock_quantity_history" />
+        <field name="arch" type="xml">
+            <field name="inventory_datetime" position="before">
+                <field
+                    name="location_id"
+                    invisible="context.get('active_model') != 'stock.valuation.layer'"
```
chienandalu, line 15: Restrict this view to the `stock.group_stock_multi_locations` group

chienandalu (commented):
I didn't success into make it work on a runbot.

- I created a new location.
- Then a received some stock into it
- When I go to the Inventory Valuation report, I can see all the valuation layers by default.
- Then I try to filter by my new location, but I see no effect at all. The whole valuation layers are still shown.

victoralmau:
Thanks @chienandalu, changes done.

chienandalu (approved):
Great! tested :+1:

PR #148, opened 2021-05-05
[13.0][IMP] stock_quantity_history_location: Add functions in test to allow re-use by othher addons

victoralmau:
Add functions in test to allow re-use by othher addons.

Related to https://github.com/OCA/stock-logistics-reporting/pull/146.

Please @chienandalu and @joao-p-marques can you review it?

@Tecnativa TT26940

### DIFF
--- a/stock_quantity_history_location/tests/test_stock_quantity_history_location.py
+++ b/stock_quantity_history_location/tests/test_stock_quantity_history_location.py
@@ -1,13 +1,16 @@
 # Copyright 2019 ForgeFlow S.L.
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo.tests.common import SavepointCase
+from odoo import fields
+
+from .common import TestCommon
 
 
-class TestStockQuantityHistoryLocation(SavepointCase):
+class TestStockQuantityHistoryLocation(TestCommon):
     @classmethod
     def setUpClass(cls):
-        super(TestStockQuantityHistoryLocation, cls).setUpClass()
+        super().setUpClass()
         cls.supplier_location = cls.env.ref("stock.stock_location_suppliers")
         cls.main_company = cls.env.ref("base.main_company")
         cls.product = cls.env.ref("product.product_product_3")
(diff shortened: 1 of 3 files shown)


chienandalu (commented):
If you're reusing the class, you should split the setup class and the actual tests one so those aren't run as many times as the class is used

stock_quantity_history_location/tests/test_stock_quantity_history_location.py
```diff
@@ -26,30 +29,32 @@ def setUpClass(cls):
...
-        move.date = "2019-08-11"
+        return move
 
     def test_wizard_past_date(self):
         wizard = self.env["stock.quantity.history"].create(
             {
                 "location_id": self.test_stock_loc.id,
                 "include_child_locations": True,
-                "inventory_datetime": "2019-08-12 00:00:00",
+                "inventory_datetime": datetime.today(),
```
rousseldenis: fields.Datetime ?

chienandalu (approved):
You didn't neet to split it into two files just into two classes (it doesn't matter if the file is the same).  

Ok if tests go :+1:

joao-p-marques (approved):

stock_quantity_history_location/wizards/stock_quantity_history.py
```diff
@@ -15,6 +18,8 @@ class StockQuantityHistory(models.TransientModel):
     def open_at_date(self):
         action = super(StockQuantityHistory, self).open_at_date()
         ctx = action["context"]
+        if isinstance(ctx, str):
+            ctx = ast.literal_eval(ctx)
```
pedrobaeza, line 22: Why are you doing this? Please use `safe_eval` if needed, but this PR message doesn't correspond to this.

pedrobaeza (changes requested):

stock_quantity_history_location/wizards/stock_quantity_history.py
victoralmau: OK sorry, I have separated this and added it to another commit and `safe_eval` does not work correctly in this case.

stock_quantity_history_location/tests/test_stock_quantity_history_location.py
```diff
@@ -26,30 +29,14 @@ def setUpClass(cls):
...
-        move.date = "2019-08-11"
+        cls._create_stock_move(cls, location_dest_id=cls.child_test_stock_loc, qty=100)
 
     def test_wizard_past_date(self):
         wizard = self.env["stock.quantity.history"].create(
             {
                 "location_id": self.test_stock_loc.id,
                 "include_child_locations": True,
-                "inventory_datetime": "2019-08-12 00:00:00",
+                "inventory_datetime": fields.Datetime.now(),
```
pedrobaeza, line 39: Why are you putting again dates as dynamic? Please let it fixed.

victoralmau: These change are necessary according to re-define code in `_create_stock_move` function and removed date in `stock.move`.
It's not correct to set date because when done move, will be auto-create `stock.valuation.layer` record, and wizard process use these code: https://github.com/odoo/odoo/blob/13.0/addons/stock_account/models/product.py#L108-L110
Too, other addon (`stock_account_quantity_history_location`) depends to these test and it's necessary to work fine.

pedrobaeza: And aren't you able to pass such date in the context?

victoralmau: IMO is not possible because `open_at_date` function https://github.com/odoo/odoo/blob/13.0/addons/stock/wizard/stock_quantity_history.py#L35) use `inventory_datetime` field.

pedrobaeza: And why aren't you forcing such date there? I'm trying to avoid random fails due to dates (or the timezone effect, where a test made around midnight fails).

victoralmau: Because `create_date`  field is auto-fill but if you prefer, I revert this change and in another PR (https://github.com/OCA/stock-logistics-reporting/pull/146) it will be necessary to do it to prevent the tests from failing.

pedrobaeza (approved):

pedrobaeza:
OK, I don't want to block more, but let's see if this brings more difficulties in the future:

/ocabot merge patch

PR #158, opened 2021-07-13
[11.0][IMP] stock_account_valuation_report: several improvements

AaronHForgeFlow:
Date in stock history wizard:

- The time on the filters is correct, but the action shows wrong dates and it confuses users

Added new filters:

- Valuation discrepancy
- Qty discrepancy


AaronHForgeFlow:
cc @ForgeFlow

PR #160, opened 2021-07-15
[FIX] [14.0] stock_card_report error creating worksheet using variants

andreampiovesana:
fix error creating worksheet using variants with the same name


marcelofrare (approved):
LGFM

PR #172, opened 2021-11-23
[12.0][FIX] stock_picking_report_valued_sale_mrp: multipicking

chienandalu:
When the recordset comes from multiple picking we could get a wrong
calculation of the mrp lines in the report when the conditions of same
component for the same order line would apply. So for every picking we
have to compute the kits valuations independently.

cc @Tecnativa TT33043

ping @victoralmau @pedrobaeza


pedrobaeza (approved):

pedrobaeza:
I fast-track the bugfix for fixing the problem.

/ocabot merge patch

Is this needed for v13?

PR #174, opened 2021-12-07
[12.0][IMP] stock_account_valuation_report

ChrisOForgeFlow:
Allows you to search for discrepancies between the inventory and accounting valuations.


rousseldenis:
@ChrisOForgeFlow Thanks for this. I think tests would be welcome.

PR #179, opened 2022-02-03
[13.0][FIX] stock_picking_report_valued_sale_mrp: multipicking

chienandalu:
Forward of #172 

When the recordset comes from multiple picking we could get a wrong calculation of the mrp lines in the report when the conditions of same component for the same order line would apply. So for every picking we have to compute the kits valuations independently.

cc @Tecnativa TT34194 TT33043

ping @victoralmau @pedrobaeza

### DIFF
--- a/stock_picking_report_valued_sale_mrp/models/stock_move_line.py
+++ b/stock_picking_report_valued_sale_mrp/models/stock_move_line.py
@@ -37,13 +37,22 @@ class StockMoveLine(models.Model):
         amounts according to the corresponding delivered kits"""
         super()._compute_sale_order_line_fields()
         pickings = self.mapped("picking_id")
-        kit_lines = pickings.move_line_ids.filtered("phantom_product_id")
         pickings.move_line_ids.update(
             {"phantom_line": False, "phantom_delivered_qty": 0.0}
         )
+        for picking in pickings:
+            self.filtered(
+                lambda x: x.picking_id == picking
+            )._compute_sale_order_line_fields_by_picking()
+
+    def _compute_sale_order_line_fields_by_picking(self):
+        """We want to compute the lines value by picking to avoid mixing lines
+        if they weren't shipped altogether.
+        """
+        kit_lines = self.filtered("phantom_product_id")
         for sale_line in kit_lines.mapped("sale_line"):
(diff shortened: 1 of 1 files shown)


stock_picking_report_valued_sale_mrp/models/stock_move_line.py
```diff
@@ -37,10 +37,18 @@ def _compute_sale_order_line_fields(self):
...
+                lambda x: x.picking_id == picking
+            )._compute_sale_order_line_fields_by_picking()
+
+    def _compute_sale_order_line_fields_by_picking(self):
+        """We want to compute the lines value by picking to avoid mixing lines
+        if they wen't altogether"""
+        kit_lines = self.filtered("phantom_product_id")
         for sale_line in kit_lines.mapped("sale_line"):
             move_lines = kit_lines.filtered(lambda x: x.sale_line == sale_line)
             # Deduct the kit quantity from the first component in the picking.
```
pedrobaeza:
```suggestion
            # Deduce the kit quantity from the first component in the picking.
```

stock_picking_report_valued_sale_mrp/models/stock_move_line.py
```diff
@@ -37,10 +37,18 @@ def _compute_sale_order_line_fields(self):
...
             {"phantom_line": False, "phantom_delivered_qty": 0.0}
         )
+        for picking in pickings:
+            self.filtered(
+                lambda x: x.picking_id == picking
+            )._compute_sale_order_line_fields_by_picking()
+
+    def _compute_sale_order_line_fields_by_picking(self):
+        """We want to compute the lines value by picking to avoid mixing lines
+        if they wen't altogether"""
```
pedrobaeza: Not sure if this is what you want to say:

```suggestion
        if they weren't shipped altogether.
        """
```

pedrobaeza (changes requested):
I miss clear steps to reproduce the problem, for example: do this and that, and you get this problem/error, and if so, reflect such steps in a regression test.

pedrobaeza:
@chienandalu can you please add the regression test?

rousseldenis:
@chienandalu Could you attend @pedrobaeza comments ?

pedrobaeza (approved):
Let's merge this for not losing the fix:

/ocabot merge patch

PR #185, opened 2022-03-18
[13.0][FIX] stock_account_quantity_history_location: location filter

chienandalu:
When filtering svl by location we don't want to let outgoing svls out.
Normally, we want all the location's related movements to get the whole
picture.

I tried to also tried to use the `product.product._get_domain_locations()` method to get the child locations when those were active, but that aproach performs poorly. Probably we should hide that option in the wizard as it will be missleading.

cc @Tecnativa TT35218

What do you think @carlosdauden @victoralmau @pedrobaeza ?

### DIFF
--- a/stock_account_quantity_history_location/wizards/stock_quantity_history.py
+++ b/stock_account_quantity_history_location/wizards/stock_quantity_history.py
@@ -29,10 +29,23 @@ class StockQuantityHistory(models.TransientModel):
             )
             action["context"] = ctx
             if self.env.context.get("active_model") == "stock.valuation.layer":
+                operator = "child_of" if self.include_child_locations else "="
                 action["domain"] = expression.AND(
                     [
                         action["domain"],
-                        [("stock_move_id.location_dest_id", "=", self.location_id.id)],
+                        [
+                            "|",
+                            (
+                                "stock_move_id.location_dest_id",
+                                operator,
+                                self.location_id.id,
+                            ),
+                            (
+                                "stock_move_id.location_id",
+                                operator,
+                                self.location_id.id,
+                            ),
+                        ],
                     ]
                 )
         else:


stock_account_quantity_history_location/wizards/stock_quantity_history.py
```diff
@@ -29,10 +37,27 @@ def open_at_date(self):
             )
             action["context"] = ctx
             if self.env.context.get("active_model") == "stock.valuation.layer":
+                # _domain_quant, domain_in, domain_out = (
+                #     self.env["product.product"]
+                #     .with_context(ctx)._get_domain_locations()
+                # )
+                # svl_domain = expression.OR(
+                #   [convert_domain(domain_in), convert_domain(domain_out)]
```
chienandalu: This was the original approach. When converting that domain to filter the related moves, the performance drops dramatically.

stock_account_quantity_history_location/wizards/stock_quantity_history.py
```diff
@@ -12,6 +12,14 @@ class StockQuantityHistory(models.TransientModel):
     _inherit = "stock.quantity.history"
 
     def open_at_date(self):
+        def convert_domain(domain):
```
pedrobaeza: For what is this needed?

chienandalu: For the alternative but non performant approach I was trying to use `product.product._get_domain_locations()`. I needed to add an extra level for using it in the stock valuation layers.

A quick fix would be to renounce to child locations in this context and add the simple domain as shown

CarlosRoca13 (approved):
LGTM

victoralmau (approved):

pedrobaeza:
/ocabot merge patch

See if needed in v14

PR #187, opened 2022-03-30
Updated from template - Switch to Github actions

rousseldenis:


pedrobaeza:
When switching to GH actions, this can't be done with the bot. It should be merged manually.

PR #189, opened 2022-04-01
[14.0][ADD] stock_report_quantity_by_location_by_date

rousseldenis:


apineux (approved):

stock_report_quantity_by_location_by_date/wizards/report_stock_quantity_location_date.xml
```diff
@@ -0,0 +1,64 @@
...
+        <field name="model">report.stock.quantity.location.date</field>
+        <field name="arch" type="xml">
+            <search string="Search Stock Report Quantity by Location">
+                <field name="product_id" />
+                <field name="company_id" />
+                <field name="location_id" groups="stock.group_stock_multi_locations" />
+                <filter
+                    name="quantity_gt_zero"
+                    string="Quantity > 0"
+                    domain="[('quantity', '>', '0.0')]"
```
acsonefho, line 49: Should be: `domain="[('quantity', '>', 0.0)]"` ?
`float` instead of `str`

stock_report_quantity_by_location_by_date/wizards/stock_report_quantity_location_date.py
```diff
@@ -0,0 +1,97 @@
...
+            pivot_date=fields.Date.to_string(self.pivot_date),
+        )
+        return "{}{}".format(report_query, added_query)
+
+    def _get_query(self):
+        report_query = ""
+        report_query = self._add_stock_data_query(report_query)
+        return report_query
+
+    def _compute_stock_report_by_location_by_date(self):
```
acsonefho, line 65: I think the `_compute` let's think about a computed field. IMO it's better to rename.

acsonefho (changes requested):

PR #197, opened 2022-05-30
14.0 stock_picking_report_valued

i-vyshnevska:
some small improvements plus make printing valuated part only for outgoing docs as picks for picking type is not available as internal moves don't have a link on sale order line and cannot compute prices correctly

### DIFF
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -4,8 +4,11 @@
 # Copyright 2016-2022 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from functools import partial
+
 from odoo import fields, models
 from odoo.tools import float_compare
+from odoo.tools.misc import formatLang
 
 
 class StockMoveLine(models.Model):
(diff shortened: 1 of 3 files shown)


stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -78,3 +81,50 @@ def _compute_sale_order_line_fields(self):
...
+    def _get_aggregated_product_quantities(self, **kwargs):
+        result = super()._get_aggregated_product_quantities(**kwargs)
+        if self.env.context.get("bypass_modification"):
+            return result
+        result = self._get_aggregated_product_quantities_delivery_price(result)
+        return result
+
+    def _get_aggregated_product_quantities_delivery_price(self, aggregated_move_lines):
+        # sale order can have sold products under different prices
+        # currently lines on stock.picking will conceder only first price
```
ajaniszewska-dev:
```suggestion
        # currently lines on stock.picking will consider only first price
```

ajaniszewska-dev (approved):

i-vyshnevska:
@rafaelbn maybe your review on this PR?

stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -78,3 +81,50 @@ def _compute_sale_order_line_fields(self):
                     "sale_price_total": valued_line.price_total,
                 }
             )
+
+    def _get_aggregated_product_quantities(self, **kwargs):
+        result = super()._get_aggregated_product_quantities(**kwargs)
+        if self.env.context.get("bypass_modification"):
```
mmequignon: I have the feeling that this context key could be more explicit

stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -78,3 +81,50 @@ def _compute_sale_order_line_fields(self):
...
+            return result
+        result = self._get_aggregated_product_quantities_delivery_price(result)
+        return result
+
+    def _get_aggregated_product_quantities_delivery_price(self, aggregated_move_lines):
+        # sale order can have sold products under different prices
+        # currently lines on stock.picking will conceder only first price
+        # so total on sale.order and stock.picking will differ
+        # consider riding of aggregating of lines for stock.picking
```
mmequignon: Should we do smth about this, like grouping by `(product, price_unit)`?

mmequignon (approved):
code looks good

stock_picking_report_valued/models/stock_move_line.py
i-vyshnevska: currently, I am not confident that I hadn't overlooked smth and maybe there is a way to compute it correctly, but that doesn't solve the possible issues with rounding in case of partials deliveries

pedrobaeza:
@dreispt please don't merge with such dirty commit history, including fixup commits.

pedrobaeza:
And this is commenting certain parts of the report, like the packages. Why? In any case, they should be removed, but after discussion. I'm reverting this meanwhile.

PR #199, opened 2022-06-22
[14.0][FIX] stock_picking_report_valued: Fix crash when you merge one picking with sale line an other without it.

sergio-teruel:
cc @Tecnativa TT37674

ping @carlosdauden @CarlosRoca13


carlosdauden (approved):

stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -55,7 +55,7 @@ def _compute_sale_order_line_fields(self):
             quantity = line._get_report_valued_quantity()
```
cesar-tecnativa, line 55: if not line.sale_line: continue

you don't have to calculate this.
quantity = line._get_report_valued_quantity()

pedrobaeza:
Already fixed in #222

PR #201, opened 2022-06-23
[FIX] stock_picking_report_valued: avoid compute when no sale line

chienandalu:
There could be some cases where the compute gets triggered but no sale
line is present. We should avoid them as we could get errors.

cc @Tecnativa TT37704

please check it @carlosdauden @sergio-teruel

### DIFF
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -51,7 +51,7 @@ class StockMoveLine(models.Model):
         access to sales orders (stricter warehouse users, inter-company
         records...).
         """
-        for line in self:
+        for line in self.filtered("sale_line"):
             quantity = line._get_report_valued_quantity()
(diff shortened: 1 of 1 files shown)


sergio-teruel:
@chienandalu look at this PR: https://github.com/OCA/stock-logistics-reporting/pull/198

chienandalu:
> @chienandalu look at this PR: #198

Yes, I'll close this one :+1:

stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -51,7 +51,7 @@ def _compute_sale_order_line_fields(self):
         access to sales orders (stricter warehouse users, inter-company
         records...).
         """
-        for line in self:
+        for line in self.filtered("sale_line"):
```
carlosdauden, line 54: You must init all records with any value to avoid cachemissed error

PR #207, opened 2022-08-03
[15.0][FIX] stock_account_valuation_report: exclude JE that are not posted

AaronHForgeFlow:
Currenctly, draft and cancelled journal items make this report incorrect because the stock value is only counted when the journal entry is posted, indeed the stock_account module only posts inventory valuation JE when posting.

cc @ForgeFlow


JordiMForgeFlow (approved):
Makes sense 👍🏼

JaumeBforgeFlow (approved):

MiquelRForgeFlow (approved):

PR #208, opened 2022-08-03
[14.0][FIX] stock_account_quantity_history_location: remove domains

JordiBForgeFlow:
Remove domains introduced in f4f1466973032a824a6951c1e1c8fe8999ab52cf
that make the valuation incorrect.

@victoralmau you introduce those changes. But we are having trouble due to those, as the inventory valuation shown is wrong, as a result of it. Layers with quantity 0 are normal (in case of landed costs, FIFO vacuum,..), and they shall be included in the report. They are included in the standard valuation report. Why should they be excluded here?

### DIFF
--- a/stock_account_quantity_history_location/__manifest__.py
+++ b/stock_account_quantity_history_location/__manifest__.py
@@ -10,7 +10,7 @@
         modules""",
     "version": "14.0.1.0.0",
     "license": "AGPL-3",
-    "author": "Eficent, Odoo Community Association (OCA)",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-reporting",
     "depends": ["stock_account", "stock_quantity_history_location"],
     "data": ["wizards/stock_quantity_history.xml"],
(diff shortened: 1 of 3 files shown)


victoralmau:
Hello, reviewing the change https://github.com/OCA/stock-logistics-reporting/commit/f4f1466973032a824a6951c1e1c8fe8999ab52cf it was only adapted to v13 and `stock.valutation.layer` the already existing functionality from v12 added in https://github.com/OCA/stock-logistics-reporting/commit/be7c68dbe52bea8e7e17b65bf15257279ec0e899.

In my opinion, I do not fully understand the need to show records that do not contribute anything because the quantity is 0 and the total price is equal to 0. Can you indicate an example in which it is necessary to show them and/or it affects the total value be wrong?

Another possibility would be to add a boolean field (hide by default) to hide or not the moves with a quantity of 0 and thus not change the current behavior showing records that others do not need.

I am also seeing that this change https://github.com/OCA/stock-logistics-reporting/commit/fd63f6f1288abbe4048e80046881b6888dc2db85 has not been carried over to v14.

Ping @chienandalu

pedrobaeza:
Closing in favor of #214

PR #212, opened 2022-08-25
[13.0][FIX] stock_account_quantity_history_location: Remove domains.

victoralmau:
Remove domains introduced in https://github.com/OCA/stock-logistics-reporting/commit/f4f1466973032a824a6951c1e1c8fe8999ab52cf that make the valuation incorrect.

Related to https://github.com/OCA/stock-logistics-reporting/pull/208

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa

PR #214, opened 2022-08-25
[14.0][FIX] stock_account_quantity_history_location: 2 fw-ports

pedrobaeza:
- Remove domains introduced in https://github.com/OCA/stock-logistics-reporting/commit/f4f1466973032a824a6951c1e1c8fe8999ab52cf that make the valuation incorrect.
- When filtering svl by location we don't want to let outgoing svls out. Normally, we want all the location's related movements to get the whole picture.

FW-ports of #185 and #212

@Tecnativa


pedrobaeza:
Fast-tracking for solving both problems:

/ocabot merge patch

PR #217, opened 2022-09-01
[15.0][FIX] stock_picking_report_valued: Calculate price unit correctly con…

ao-landoo:
…sidering uom

We fixed the calculation error of price unit on the stock picking report valued.

Before: Directly got value of price unit without looking at the unit of measure on the sale order.

Now: It takes into account the conversion ratio between the unit of measure on the sale order and the unit of measure on the stock picking in order to get the real price unit.

### DIFF
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -20,7 +20,8 @@ class StockMoveLine(models.Model):
         related="sale_line.tax_id", readonly=True, string="Sale Tax"
     )
     sale_price_unit = fields.Float(
-        related="sale_line.price_unit", readonly=True, string="Sale price unit"
+        compute="_compute_sale_order_line_fields",
+        readonly=True,
     )
     sale_discount = fields.Float(
         related="sale_line.discount", readonly=True, string="Sale discount (%)"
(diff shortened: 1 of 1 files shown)


bi-landoo (approved):
Functional Approval

bi-landoo:
@pedrobaeza @ernestotejeda  ¿Could you review this fix please?

stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -49,11 +50,16 @@ def _compute_sale_order_line_fields(self):
         """
         for line in self:
             sale_line = line.sale_line
+            sale_line_uom = sale_line.product_uom
             price_unit = (
                 sale_line.price_subtotal / sale_line.product_uom_qty
                 if sale_line.product_uom_qty
                 else sale_line.price_reduce
             )
+            if sale_line_uom.uom_type == "bigger":
```
pedrobaeza: Please use `uom.uom` `convert` method for changing the quantity in the previous expression.

ao-landoo: I used `_compute_price` method as I found no `convert` method

stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -49,11 +50,13 @@ def _compute_sale_order_line_fields(self):
         """
         for line in self:
             sale_line = line.sale_line
+            sale_line_uom = sale_line.product_uom
             price_unit = (
                 sale_line.price_subtotal / sale_line.product_uom_qty
                 if sale_line.product_uom_qty
                 else sale_line.price_reduce
             )
+            price_unit = sale_line_uom._compute_price(price_unit, line.product_uom_id)
```
pedrobaeza:
```suggestion
            if sale_line_uom != line.product_uom_id:
                price_unit = sale_line_uom._compute_price(price_unit, line.product_uom_id)
```

pedrobaeza (approved):

PR #219, opened 2022-09-02
[13.0][FIX] stock_picking_report_valued: Calculate price unit correctly con…

ao-landoo:
…sidering uom

Backport of #218 

We fixed the calculation error of price unit on the stock picking report valued.

Before: Directly got value of price unit without looking at the unit of measure on the sale order.

Now: It takes into account the conversion ratio between the unit of measure on the sale order and the unit of measure on the stock picking in order to get the real price unit.


ao-landoo:
@pedrobaeza Can it be merged? Thanks

pedrobaeza (approved):

PR #221, opened 2022-09-07
[13.0][FIX] stock_picking_report_valued: valued line computation

chienandalu:
The fix in fac3997ee518f2c984776d8a06860fb5e80ccee2 didn't consider move lines without a linked sale line. That would end up in inconsistent calculations and even the original line rewriting.

cc @Tecnativa TT38863

ping @pedrobaeza @carlosdauden 

cc @ao-landoo

### DIFF
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -53,25 +53,35 @@ class StockMoveLine(models.Model):
         access to sales orders (stricter warehouse users, inter-company
         records...).
         """
+        self.sale_tax_description = False
+        self.sale_price_subtotal = False
+        self.sale_price_tax = False
+        self.sale_price_total = False
+        self.sale_price_unit = False
         for line in self:
-            quantity = line._get_report_valued_quantity()
             valued_line = line.sale_line
+            if not valued_line:
+                continue
+            quantity = line._get_report_valued_quantity()
             sale_line_uom = valued_line.product_uom
+            different_uom = valued_line.product_uom != line.product_uom_id
             # If order line quantity don't match with move line quantity compute values
-            if valued_line and float_compare(
+            different_qty = float_compare(
                 quantity,
                 line.sale_line.product_uom_qty,
                 precision_rounding=line.product_uom_id.rounding,
-            ):
+            )
+            if different_uom or different_qty:
                 # Force read to cache M2M field for get values with _convert_to_write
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I see that this is not enough: we need to do other `if` level. The first one is `if valued_line`, and the second one if the quantity is different, and in the first level, do the UoM check.

chienandalu:
Changes done:

- Values of the recordset initialized at first
-  We exit the loop if no linked sale line.
- Create the virtual line with the copy if we have to recompute the prices either for qty or uom
- Update price_unit accordingly depending on the case

stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -53,25 +53,35 @@ def _compute_sale_order_line_fields(self):
...
                 line.sale_line.mapped("tax_id")
                 # Create virtual sale line with stock move line quantity
                 sol_vals = line.sale_line._convert_to_write(line.sale_line._cache)
                 valued_line = line.sale_line.new(sol_vals)
                 valued_line.product_uom_qty = quantity
+            if different_qty:
                 # Force original price unit to avoid pricelist recomputed (not needed)
                 valued_line.price_unit = line.sale_line.price_unit
-            if sale_line_uom != line.product_uom_id:
+            if different_uom:
```
pedrobaeza, line 84: This is still overwriting the sale line price.

chienandalu: Look at line 79

pedrobaeza: But we don't need to create a whole new line just for that conversion. Better to store the price unit in a variable.

chienandalu: Well, is following the previous implementation that triggers the line totals with the new quantity or the new price unit but in a virtual line: https://github.com/OCA/OCB/blob/745b5dad8c7bf817a7d60076c1d12307132e71b1/addons/sale/models/sale.py#L1088

pedrobaeza (approved):

sergio-teruel (approved):

pedrobaeza:
@ao-landoo please check that your PR was causing this side effect.

pedrobaeza:
Let's fast-track it for voiding further problems:

/ocabot merge patch

pedrobaeza:
This should be fw-port it to v14 and v15. @ao-landoo do you handle it?

PR #229, opened 2022-09-28
[15.0][IMP] stock_account_valuation_report: Add checkbox to include Draft E…

BernatPForgeFlow:
…ntries in Dual Inventory Valuation

### DIFF
--- a/stock_account_valuation_report/models/product_product.py
+++ b/stock_account_valuation_report/models/product_product.py
@@ -67,17 +67,24 @@ class ProductProduct(models.Model):
     def _compute_inventory_value(self):
         self.env["account.move.line"].check_access_rights("read")
         to_date = self.env.context.get("at_date", False)
+        include_draft_entries = self.env.context.get("include_draft_entries", False)
         accounting_values = {}
         layer_values = {}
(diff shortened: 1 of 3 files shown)


CasVissers-360ERP (approved):
Functional review

stock_account_valuation_report/models/product_product.py
```diff
@@ -67,17 +67,24 @@ def _search_valuation_discrepancy(self, operator, value):
...
-        query = """
+        query = (
+            """
             SELECT aml.product_id, aml.account_id,
             sum(aml.balance), sum(quantity),
             array_agg(aml.id)
             FROM account_move_line AS aml
+            INNER JOIN account_move AS am ON am.id = aml.move_id
             WHERE aml.product_id IN %%s
+            AND am.state """
```
ghost, line 82: Same comment as the one on the PR #230 

Is it necessary to do a JOIN on this ? Isn't it enough to use the parent_state field ?
https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_move.py#L3637

ghost (changes requested):

PR #243, opened 2022-11-09
[14.0][IMP] stock_report_quantity_by_location: add virtual_available qty

enriquemartin:


stock_report_quantity_by_location/wizards/stock_report_quantity_by_location.py
```diff
@@ -69,6 +73,7 @@ def _compute_stock_report_by_location(self):
                             "quantity_on_hand": qty_on_hand,
                             "quantity_reserved": qty_reserved,
                             "quantity_unreserved": qty_unreserved,
+                            "quantity_available": product.virtual_available,
```
rousseldenis, line 76: @enriquemartin Why not adding the field to the read_group ? With this, you will impact performances as qty_available function will be called twice

rousseldenis (changes requested):
IMHO needs fixing

stock_report_quantity_by_location/wizards/stock_report_quantity_by_location.py
enriquemartin: The read_group is executed on `stock.quant`, and the `virtual_available` is computed on product.
In fact the `available_quantity` of the `stock.quant` can't have direct relationship with the `virtual_available` of the product as the `stock.quant` can't be reserved if there is no stock.
That's the main reason of this PR: having the `virtual_available` in the report to filter the products with negative availability.

enriquemartin: Have I answered your doubts @rousseldenis?

PR #244, opened 2022-11-10
[15.0][FIX] stock_picking_report_valued: Display all picking types reports …

ao-landoo:
…correcty

Displays all picking type reports correctly, before this fix, incoming type picking report was broken

### DIFF
--- a/stock_picking_report_valued/report/stock_picking_report_valued.xml
+++ b/stock_picking_report_valued/report/stock_picking_report_valued.xml
@@ -2,11 +2,14 @@
 <odoo>
 
     <template id="valued_report_picking" inherit_id="stock.report_delivery_document">
+        <xpath expr="//div[hasclass('page')]" position="before">
+            <t t-set="is_outgoing" t-value="o.picking_type_code == 'outgoing'" />
+        </xpath>
         <xpath
             expr="//table[@t-if=&quot;o.move_line_ids and o.state==&apos;done&apos;&quot;]/thead/tr"
             position="inside"
         >
-            <t t-if="o.valued and o.sale_id and o.move_line_ids">
+            <t t-if="o.valued and o.sale_id and o.move_line_ids and is_outgoing">
                 <t t-if="o.state != 'done'">
                     <th class="text-right"><strong>Qty Reserved</strong></th>
                 </t>
@@ -21,7 +24,7 @@
         <xpath expr="//th[@name='th_sml_qty_ordered']" position="attributes">
             <attribute
                 name="t-if"
-                add="and not (o.valued or o.sale_id or o.move_line_ids)"
+                add="and (not is_outgoing or not (o.valued and o.sale_id and o.move_line_ids))"
                 separator=" "
             />
         </xpath>
@@ -31,7 +34,7 @@
         >
             <attribute
                 name="t-if"
-                add="or (o.valued and o.sale_id and o.move_line_ids)"
+                add="or (o.valued and o.sale_id and o.move_line_ids and is_outgoing)"
                 separator=" "
             />
         </xpath>
@@ -41,7 +44,7 @@
         >
             <attribute
                 name="t-if"
-                add="or (o.valued and o.sale_id and o.move_line_ids)"
+                add="or (o.valued and o.sale_id and o.move_line_ids and is_outgoing)"
                 separator=" "
             />
         </xpath>
@@ -51,7 +54,7 @@
         >
             <attribute
                 name="t-if"
-                add="or (o.valued and o.sale_id and o.move_line_ids)"
+                add="or (o.valued and o.sale_id and o.move_line_ids and is_outgoing)"
                 separator=" "
             />
         </xpath>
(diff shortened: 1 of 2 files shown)


bi-landoo (approved):
Tried all combinations and works perfectly

bi-landoo:
@pedrobaeza @ao-landoo  @sergio-teruel @ernestotejeda could you please test it and if all correct, give approval.

Thanks!

ma-landoo (approved):

pedrobaeza (commented):
Instead of repeating each time `o.picking_type_code == 'outgoing'`, put a `<t t-set="is_outgoing" t-value="o.picking_type_code == 'outgoing"/>`, and then use it for simplifying the expressions.

ao-landoo:
@pedrobaeza pre-commit error doesn't seem to be related to this PR

pedrobaeza:
Please update copier template in a separate PR.

ao-landoo:
@pedrobaeza Can it be merged?

stock_picking_report_valued/report/stock_picking_report_valued.xml
```diff
@@ -91,8 +94,7 @@
         inherit_id="stock.stock_report_delivery_has_serial_move_line"
     >
         <xpath expr="//td[@name='move_line_lot_qty_done']" position="after">
-            <t t-if="o.valued and o.sale_id and o.move_line_ids">
-                <t t-if="move_line.picking_id.state != 'done'">
+            <t t-if="move_line.picking_id.state != 'done'">
```
pedrobaeza, line 97: This is changing the logic of the report.

pedrobaeza (changes requested):

stock_picking_report_valued/report/stock_picking_report_valued.xml
ao-landoo: @pedrobaeza 
Quantity fields' visibility was conditioned by `o.valued`, `o.sale_id` and `o.move_line_ids`, which makes no sense. These fields should always be displayed, not only if the picking fulfills those conditions.

I modified this so that the only columns affected by these conditions are price unit, discount, subtotal, etc.

pedrobaeza: They should be there for a reason. @carlosdauden can you tell us why they are there?

carlosdauden: The original idea of this module was to allow printing a valued delivery note in the following cases:
- picking not reserved (move_line_ids is empty, get lines from move_lines field product_uom_qty)
- picking reserved, but not done (get quantity from move_line_ids field qty_done or product_uom_qty)
- picking done (get quantity from move_line_ids field product_uom_qty)

If these premises are met, welcome any improvement

ao-landoo: What do you mean with original idea? That was the initial plan of the module?

If i'm not wrong the module is not working the way you describe it right now on branch 15.0 (without this PR).

pedrobaeza: Then what Carlos describes is complied with these changes, Aritz?

carlosdauden: True, the module does not work as described at this time in the 15.0 branch.
Since it is modified, would you be so kind as to adjust it to the desired behavior? (or as similar as possible)

js-landoo: Hi @carlosdauden  and @pedrobaeza 
As @ao-landoo and @bi-landoo are in holidays this week, I will try to help with this PR.

To simplify the topic, In this PR we will cover only features of outcoming stock pickings.

After deep testing 14 different use cases In last 15.0 OCA version, we only find one small bug.
Bug: "Ordered" label is not printed
Conditions to reproduce: from stock picking in state "Done" and valued "NO", print a delivery slip
image
Today Runboat seems not working so you can test it in our database
https://demo15e.landoo.es/web?db=demo15e_bi_sprv
admin-oca / admin-oca

With PR#244 you can see previous bug is not present and all desired behaviour works as original module, in 3 different states:
1) picking not reserved (move_line_ids is empty, get lines from move_lines field product_uom_qty)
2) picking reserved, but not done (get quantity from move_line_ids field qty_done or product_uom_qty)
3) picking done (get quantity from move_line_ids field product_uom_qty)
You can test in
https://demo15e.landoo.es/web?db=demo15e_bi_sprv_landoo
admin-oca / admin-oca
In this DB you can see module renamed as stock_picking_report_valued_landoo 
image

Also we are working to use this module with all features and incoming stock pickings.
Once we test all use cases, we will upload a new PR in next weeks.

I wait for your comments. Many thanks.

pedrobaeza (approved):
Ok, let's merged it. I take your promise of checking all the exposed use cases following weeks.

/ocabot merge minor

js-landoo:
Thanks @pedrobaeza 
All exposed use cases are already done. I will send you a private message with our complete test battery.

What we are going to test in next weeks are all INCOMING use cases.
I understood this module was not developed with this objectiv and that's why it's not working fine here.
We need that for some customers, so we are improving that and code changes and tests have a bigger effort. That's why we need more time.
Regards.

pedrobaeza:
Ok, no need for such message. I trust you.

PR #252, opened 2022-12-13
[14.0][FIX] stock_picking_report_custom_description: Transfer line description to move

pedrobaeza:
Since v14, the SO line custom description is not transferred to the stock.move name, so we need to do such transfer in this module, as the main purpose of it is to print custom texts from SO.

@Tecnativa TT38786


pilarvargas-tecnativa (approved):
Approved!

stock_picking_report_custom_description/tests/test_stock_picking_report_custom_description.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2022 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestStockPickingReportCustomDescription(common.TransactionCase):
```
ernestotejeda:
```suggestion
class TestStockPickingReportCustomDescription(common.SavepointCase):
```

pedrobaeza: True, this is v14. What a pity that they renamed the other way the classes that provokes these misunderstandings...

stock_picking_report_custom_description/models/stock_rule.py
```diff
@@ -0,0 +1,35 @@
...
+            product_uom,
+            location_id,
+            name,
+            origin,
+            company_id,
+            values,
+        )
+        if values.get("sale_line_id"):
+            line = self.env["sale.order.line"].browse(values["sale_line_id"])
+            res["name"] = line.name
```
rousseldenis, line 34: IMHO, this does more than it should for a 'report' module.

It should be done in a specific module like 'sale_stock_order_line_name' or ...

pedrobaeza: The feature and the goal is the same than in previous versions, so we don't see the need of adding a new module. It's explained already on the README, so people knows what to get when they install it, and those with the module installed in previous versions will get the same functionality. The question is that this was handled by Odoo itself previously, and now it's not.

ernestotejeda (approved):

stock_picking_report_custom_description/models/stock_rule.py
rousseldenis: Yes, I understood that.

But, you mixed two functionalities in the same module. This should be avoided (as some people maybe want to get the transmission of sale line description to move without the custom description module...)

pedrobaeza: They are separated reports, so no problem on that. And they can also create another module with just that feature if they want.

rousseldenis: > And they can also create another module with just that feature if they want.

So, duplicated code

pedrobaeza: Yes, but that's only if someone like you insists :stuck_out_tongue_closed_eyes:

rousseldenis: Try to do a better OCA world. I'm a little bit utopian :-)

pedrobaeza: Well, I know that, but sometimes we need to be pragmatic, and I assure you that we already had this discussion before going this way, as the other will require a lot of more work for the migration path of our customers, and we are talking about 10 lines of code (20 with black, hehe). It's a good trade-off.

rousseldenis: Yes, let's do in this way for now but keeping this thread in mind

PR #273, opened 2023-03-31
[15.0][FIX] stock_picking_report_product_sticker: Show stickers in different positions

Shide:
Show stickers in different positions and reduce the image size on the report.

Top position:
top

Bottom position:
bottom

MT-2153 @moduon @rafaelbn @yajo please review

### DIFF
--- a/stock_picking_report_product_sticker/models/stock_picking.py
+++ b/stock_picking_report_product_sticker/models/stock_picking.py
@@ -3,15 +3,18 @@
 
 from odoo import api, fields, models
 
+from .stock_picking_type import REPORT_STICKER_POSITIONS
+
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
-    show_product_stickers = fields.Boolean(
-        help="Show Product Stickers on pickings of this type.",
+    show_product_stickers = fields.Selection(
+        selection=REPORT_STICKER_POSITIONS,
         compute="_compute_show_product_stickers",
         store=True,
         readonly=False,
+        help="Show Product Stickers on pickings of this type.",
     )
     sticker_ids = fields.Many2many(
         comodel_name="product.sticker",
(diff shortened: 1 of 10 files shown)


Shide:
Fails `stock_move_value_report` module on tests

stock_picking_report_product_sticker/models/stock_picking.py
```diff
@@ -20,7 +23,7 @@ class StockPicking(models.Model):
         store=False,
     )
 
-    @api.depends("picking_type_id.show_product_stickers")
+    @api.depends("picking_type_id")
     def _compute_show_product_stickers(self):
         for picking in self:
             picking.show_product_stickers = (
```
yajo, line 29: This compute seems like it will fail now, as it assigns a bool value.

yajo (changes requested):

stock_picking_report_product_sticker/models/stock_picking.py
Shide: I'm also refactoring the field changing type to Selection instead of Bool. Doesn't make sense to me split the functionality in 2 fields.
If you don't want to display stickers, just leave it blank on picking type or on the picking.
If you want to display stickers, choose a position (bottom or top).

And I've removed the `.show_product_stickers` to simplify

yajo: Ah yes, sorry indeed! 🤦🏼‍♂️

yajo (approved):

rafaelbn (approved):
This is more a improvement than a FIX 😄 

Really nice and needed!

👍

Shide:
Merged as major because the change on the field type

rafaelbn:
@etobella @rousseldenis please could you help with this merge? @Shide is maintainer but ocabot don't recognize it.

rousseldenis:
@Shide Could you fix tests ?

rousseldenis:
@rafaelbn @Shide The maintainers key is a plural, change that one too:

image

Shide:
@rousseldenis done! :)

Shide:
@etobella please merge if you deserve. tests and maintainer key are fixed 😄

PR #282, opened 2023-08-18
[FIX] stock_quantity_history_location: Migrate js to OWL

rolandojduartem:
This commit solves the error
"Non loaded module
stock_quantity_history_location.InventoryReportListController" because
the dependency stock.InventoryReportListController was missed. The
dependency from stock was migrated to OWL [1], so it is inherited using the
OWL to preserve the behavior.

References:
- [1] https://github.com/odoo/odoo/commit/ec0b1005


rolandojduartem:
This is in development phase

rolandojduartem:
Hi, @luisg123v, @imanie383 and @desdelinux, please, could you review this PR? Do you know anything about the lint in the pre-commit?

luisg123v:
Hi @sebasdrk17,

Could you review, please?

stock_quantity_history_location/__manifest__.py
```diff
@@ -5,21 +5,22 @@
...
     "assets": {
         "web.assets_backend": [
-            "stock_quantity_history_location/static/src/js/inventory_report.js",
-        ],
-        "web.assets_qweb": [
-            "stock_quantity_history_location/static/src/xml/inventory_report.xml",
+            "stock_quantity_history_location/static/src/components"
+            "/inventory_report/inventory_report.js",
+            "stock_quantity_history_location/static/src/components/"
+            "inventory_report//inventory_report.xml",
```
luisg123v: Extra slash

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates>
```
luisg123v: Add name

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.xml
```diff
@@ -0,0 +1,26 @@
...
+    <t t-inherit="stock.InventoryReport.Buttons" t-inherit-mode="extension" owl="1">
+        <xpath expr="//button[contains(@t-if, 'stock.quant')]" position="attributes">
+            <attribute name="t-if" add="and !this.multi_location" separator=" " />
```
luisg123v: There's no safe way to add conditions to an attribute, as order is not guaranteed

I'd suggest creating a `<t t-if` element and moving button there, and inserting the new button just after that element using `t-else`

luisg123v (changes requested):

stock_quantity_history_location/__manifest__.py
```diff
@@ -5,21 +5,22 @@
...
     "data": ["wizards/stock_quantity_history.xml"],
     "assets": {
         "web.assets_backend": [
-            "stock_quantity_history_location/static/src/js/inventory_report.js",
-        ],
-        "web.assets_qweb": [
-            "stock_quantity_history_location/static/src/xml/inventory_report.xml",
+            "stock_quantity_history_location/static/src/components"
+            "/inventory_report/inventory_report.js",
+            "stock_quantity_history_location/static/src/components/"
```
imanie383, line 22: you are not importing any files in L22

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.js
```diff
@@ -0,0 +1,18 @@
+/** @odoo-module */
+
+import session from "web.session";
+import {patch} from "@web/core/utils/patch";
+import {InventoryReportListController} from "@stock/views/list/inventory_report_list_controller";
+const {onWillStart} = owl;
+
+patch(InventoryReportListController.prototype, "inventory_report_list_controller", {
+    setup() {
+        this._super(...arguments);
```
imanie383, line 10:
```suggestion
        super.setup();
```

If it is a component, you need this super

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.js
```diff
@@ -0,0 +1,18 @@
...
+        this.multi_location = false;
+        onWillStart(async () => {
+            this.multi_location = await session.user_has_group(
```
imanie383: image
https://www.odoo.com/documentation/16.0/developer/reference/frontend/services.html#using-a-service

image

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.xml
luisg123v: I meant ID:

```suggestion
<templates id="template" xml:space="preserve">
```

imanie383 (changes requested):

stock_quantity_history_location/__manifest__.py
rolandojduartem: It was importing the file.

rolandojduartem:
Hi, @luisg123v and @imanie383, I applied your suggestions.

sebasdrk17:
@luisg123v  Am I still needed here? I've seen that @imanie383 has already review

stock_quantity_history_location/__manifest__.py
luisg123v: @imanie383 to clarify, this is a multiline string, next line contains file name.

luisg123v:
> @luisg123v Am I still needed here? I've seen that @imanie383 has already review

@sebasdrk17 maybe not as needed anymore, but any comment is welcomed anyway.

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates id="template" xml:space="preserve">
+    <t t-inherit="stock.InventoryReport.Buttons" t-inherit-mode="extension" owl="1">
+        <xpath expr="(//div/*)[last()]" position="after">
```
luisg123v: AFAIK this is inserting elements at the end, even after the "Appli All" button.

Maybe what you need is to place this before the "Apply All" button, by filtering by `t-on-click="onClickApplyAll"`.

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.xml
```diff
@@ -0,0 +1,28 @@
...
+            <t t-if="this.multi_location" name="has_multi_location">
+                <button
+                    t-if="
+                        props.resModel == 'stock.quant'
+                        and (
+                            !props.context.inventory_mode
+                            or props.context.inventory_report_mode
+                        )
+                        and !props.context.no_at_date
```
luisg123v: Add an ID to this button, so if we (or anybody else) migrate the module `stock_account_quantity_history_location`, xpaths are simpler

luisg123v (changes requested):

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.js
```diff
@@ -0,0 +1,17 @@
+/** @odoo-module */
```
imanie383:
```suggestion
/** @odoo-module **/
```

stock_quantity_history_location/__manifest__.py
sebasdrk17: **Up to you**

you could just use a shortpath

`stock_quantity_history_location/static/src/**/*.xml`

example:
Screenshot 2023-08-21 at 1 06 39 p m

rolandojduartem:
Hi, @luisg123v and @imanie383, I applied your suggestions.

stock_quantity_history_location/__manifest__.py
rolandojduartem: I know, but @luisg123v prefers explicit file paths when there are few files.

rolandojduartem:
Hi,  @luisg123v and @imanie383, could you review this PR again? CI is green now

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.xml
```diff
@@ -0,0 +1,29 @@
...
+                    class="btn btn-primary o_button_at_date me-2"
+                    t-on-click="onClickInventoryAtDate"
+                >
+                    Inventory at Date &amp; Location
+                </button>
+            </t>
+            <t t-else="" name="has_not_multi_location" />
+        </button>
+        <t name="has_not_multi_location" position="inside">
+            <xpath expr="//button[contains(@t-if, 'stock.quant')]" position="move" />
```
imanie383: image

imanie383: Wait, I'm not sure why you need another button if your button is exactly the same as the original button(if and click-method)

rolandojduartem: Actually, hasclass('o_button_at_date') is more ambiguous because stock_account add another button with that class, actually using the model is more specific for every button. What do you think? (This module will work together with stock_account_quantity_history_location if stock_account_quantity_history_location is migrated).

luisg123v: I agree with @rolandojduartem, current filter is preferred to using class.

Regarding this:

> Wait, I'm not sure why you need another button if your button is exactly the same as the original button(if and click-method)

Because the text is different and there's no way to change or hide current text.

imanie383: you can use a t-out like 

`<button t-out="'demo'">imanie</button>`

that overwrites the inner text

stock_quantity_history_location/static/src/components/inventory_report/inventory_report.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates id="template" xml:space="preserve">
+    <t t-inherit="stock.InventoryReport.Buttons" t-inherit-mode="extension" owl="1">
+        <button t-on-click="onClickApplyAll" position="before">
+            <t t-if="this.multi_location" name="has_multi_location">
```
imanie383:
```suggestion
            <t t-if="multi_location" name="has_multi_location">
```

image

rolandojduartem:
Hi, @luisg123v and @imanie383, what do you think now? I could not make the groups attribute works for this inheritance, it would be great to remove the js file

luisg123v (approved):
LGTM :+1: 

/ocabot merge nobump

Not increasing version because this change is already increasing it.

PR #288, opened 2023-09-21
[16.0][MIG] stock_picking_report_custom_description: Migration to version 16.0

carolinafernandez-tecnativa:
Migration to v16 with improvements:

- Transfer description sales order line to description picking instead of stock.move > name
- Remove customs reports as standard report print description picking field
- When sales order line changes, description picking is updated

@Tecnativa

TT45186

@pedrobaeza @pilarvargas-tecnativa


pilarvargas-tecnativa (approved):
Functionality tested locally, runboat has error not related to migration. LGTM :) thanks!

pedrobaeza:
Testing on runboat after fixing what other migrators have screwed up (PR in the way), I see that the custom text is put after the product name:

Selección_070

and I'm not sure if that's the intention, as you usually leave the product description and slightly alter it, and with this, you will have 2 texts very similar. Does this work the same way on previous versions?

Another thing: the description from the sales order line to the picking is not transferred. I think there's another module for that, isn't it?

pilarvargas-tecnativa:
> Testing on runboat after fixing what other migrators have screwed up (PR in the way), I see that the custom text is put after the product name:
> 

In my tests, the product name doesn't appear

image
image

pedrobaeza:
#289 for fixing the error in runboat.

pedrobaeza:
Please avoid to quote entire text, specially when there are images involved, as it takes a lot of vertical space in the thread, and more scroll is needed. I have edited your comment to reduce it. And moreover, your comment is next to mine, so it's very easy to be followed.

About what you see, not sure if the problem is any extra module that is putting the other text. I'm going to try.

pedrobaeza:
I have removed all the OCA modules in runboat except this one, and I get the same. Even more, I'm printing standard "Delivery slip", and I get:

Selección_071

and printing "Delivery Slip (Description)":

Selección_072

so, should we deprecate this module if standard does this in a similar way? How do you get to the situation where the product name is not shown?

As reference, this is the picking tested, which is in ready state (not sure if it can affect): http://oca-stock-logistics-reporting-16-0-pr288-3d51c290d9b0.runboat.odoo-community.org/web#id=21&cids=1&menu_id=174&action=330&active_id=2&model=stock.picking&view_type=form

pilarvargas-tecnativa:
My tests have been locally with the database completely cleaned (attached gif).
print

I have tried again on the link you provide and it fails to print "Delivery Slip (Description)", it appears as you indicate but I think it's a runboat thing.

sergio-teruel:
The module works as expected.
@pedrobaeza the text that you remark is the description for devilery settin on product.

sergio-teruel (approved):

carolinafernandez-tecnativa:
@pedrobaeza @pilarvargas-tecnativa @sergio-teruel could you please check it now? Thanks!

pedrobaeza:
Tested and working. The product display_name is still shown in the delivery slip, but I suppose that's the normal way of working and it was happening also on the "Delivery slip (description)" report.

carolinafernandez-tecnativa:
ping @pedrobaeza

stock_picking_report_custom_description/models/sale_order_line.py
```diff
@@ -0,0 +1,16 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("name"):
+            move = self.env["stock.move"].search([("sale_line_id", "=", self.id)])
+            if move:
+                move.description_picking = vals.get("name")
```
pedrobaeza: This is multi:

```suggestion
            moves = self.env["stock.move"].search([("sale_line_id", "in", self.ids)])
            if moves:
                moves.write({"description_picking": vals["name"]})
```

PR #292, opened 2023-10-17
[14.0][MIG] stock_inventory_valuation_report

solo4games:

PR #299, opened 2024-01-02
[14.0][MIG] stock_inventory_valuation_report

HekkiMelody:
Supersedes https://github.com/OCA/stock-logistics-reporting/pull/292

The last three commits add:
- Remove today's date from the report footer to avoid confusion
- The report prints the variant's name, instead of the product template's name
- the UoM in both XLSX and PDF reports


francesco-ooops:
@geomer198 could you review?

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
```diff
@@ -0,0 +1,120 @@
...
+            self.env["product.product"]
+            .with_context(
+                to_date=self.inventory_datetime,
+                company_owned=True,
+                create=False,
+                edit=False,
+            )
+            .search(domain)
+            # 'quantity_svl' we can't use compute field for search
+        ).filtered(lambda pp: pp.quantity_svl != 0)
```
geomer198, line 73: Why do you use `filtered` function, if you could add domain in search function.
```py
.search(expression.AND([domain, [('quantity_svl', '!=', 0)]]))
```
This solution is faster than using `filtered` function.

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
```diff
@@ -0,0 +1,120 @@
...
+        results = self.env["stock.inventory.valuation.view"]
+        if products:
+            for product in products:
```
geomer198, line 76: Here you could use `new` function once. In first step you prepare vals and second step `new` function.
```py
vals = [{....} for product in products]
results = resultnts.new(vals)
```

stock_inventory_valuation_report/wizard/stock_quantity_history.py
```diff
@@ -0,0 +1,46 @@
...
+            "stock_inventory_valuation_report."
+            "action_stock_inventory_valuation_report_html"
+        )
+        vals = action.read()[0]
+        new_context = vals.get("context", {})
+        if isinstance(new_context, str):
+            new_context = safe_eval(new_context)
+        model = self.env["report.stock.inventory.valuation.report"]
+        report = model.create(self._prepare_stock_inventory_valuation_report())
+        new_context["active_id"] = report.id
```
geomer198: I think better use `update` function.
```py
new_context.update(active_id=report.id, active_ids=report.ids)
```

stock_inventory_valuation_report/wizard/stock_quantity_history.py
```diff
@@ -0,0 +1,46 @@
...
+    def button_export_html(self):
+        self.ensure_one()
+        action = self.env.ref(
+            "stock_inventory_valuation_report."
+            "action_stock_inventory_valuation_report_html"
+        )
+        vals = action.read()[0]
+        new_context = vals.get("context", {})
+        if isinstance(new_context, str):
+            new_context = safe_eval(new_context)
```
geomer198: I think need add `try except` block for convert value.

stock_inventory_valuation_report/tests/test_stock_inventory_valuation_report.py
```diff
@@ -0,0 +1,221 @@
...
+            [self.report.id],
+            report_type="xlsx",
+        )
+
+    def test_print(self):
+        self.report.print_report("qweb")
+        self.report.print_report("xlsx")
+
+
+class TestStockInventoryValuationReport(common.SavepointCase):
```
geomer198, line 72: Could you add a message for each assert.

stock_inventory_valuation_report/static/src/js/stock_inventory_valuation_report_backend.js
```diff
@@ -0,0 +1,116 @@
...
+                this.given_context.ttype = action.context.ttype || false;
+            },
+            willStart: function () {
+                return Promise.all([
+                    this._super.apply(this, arguments),
+                    this.get_html(),
+                ]);
+            },
+            set_html: function () {
+                var self = this;
```
geomer198: I think use `self` variable isn't correct. For using global `this` into functions you can use this function format `() => {}`.
For example:
```js
def.then(() => {
    this.report_widget.$el.html(this.html);
});
```

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+    cost_currency_id = fields.Many2one(
+        comodel_name="res.currency",
+    )
+    standard_price = fields.Float()
+    stock_value = fields.Float()
+    cost_method = fields.Char()
+
+
+class StockInventoryValuationReport(models.TransientModel):
```
geomer198, line 32: Please add docstrings for methods.

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
```diff
@@ -0,0 +1,120 @@
...
+        comodel_name="stock.inventory.valuation.view",
+        inverse_name="report_id",
+        compute="_compute_results",
+    )
+
+    @api.depends("inventory_datetime")
+    def _compute_results(self):
+        self.ensure_one()
+        domain = [("type", "=", "product")]
+        product_id = self.env.context.get("product_id", False)
```
geomer198: By default `get` function returned `None` value, because in `if` is `None` equal to `False`.

stock_inventory_valuation_report/wizard/stock_quantity_history.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2019 Ecosoft Co., Ltd. (http://ecosoft.co.th)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+from odoo.tools.safe_eval import safe_eval
+
+
+class StockQuantityHistory(models.TransientModel):
```
geomer198, line 11: Here too. Please add docstrings for methods.

geomer198 (changes requested):
Could you please fix code by my comments.

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
HekkiMelody: as the comment states, quantity_svl is not a stored field, and cannot be used in the `search`

stock_inventory_valuation_report/wizard/stock_quantity_history.py
HekkiMelody: I added some docstrings, but I don't find it very helpful to add a docstring "Produces the HTML report" to a method that's already called "button_export_html".

Doesn't that just create more noise?

stock_inventory_valuation_report/reports/stock_inventory_valuation_report.py
HekkiMelody: `.new()` cannot take a list of dictionaries, only a single dictionary.

francesco-ooops:
@geomer198 better now?

also requesting review from @jbaudoux and @rousseldenis

geomer198 (approved):
LGTM!

francesco-ooops:
@OCA/logistics-maintainers can this be merged?

francesco-ooops (approved):
Functional is ok!

francesco-ooops:
@rousseldenis good to go?

francesco-ooops:
@pedrobaeza can it be merged based on existing reviews?

PR #300, opened 2024-01-17
[17.0][MIG] stock_picking_report_valued: Migration to 17.0

nguyenminhchien:
Ref: BSRD-711

I improve this module a bit to make it works with the column of **HS Code** (after installing the module `stock_delivery`).
It means that we don't need to migrate the module `stock_picking_report_valued_delivery` in version 17.0 anymore.

# Result:
- Report has both Lot Number and HS Code
lot_and_hscode

- Report has Lot Number but not HS Code
lot-no-hscode

- Report has HS Code but not Lot Number
hscode-no-lot-

- Report has no Lot Number nor HS Code
no-lot-no-hscode


stock_picking_report_valued/report/stock_picking_report_valued.xml
```diff
@@ -1,18 +1,16 @@
...
         </xpath>
-        <xpath
-            expr="//table[@t-if=&quot;o.move_line_ids and o.state==&apos;done&apos;&quot;]/thead/tr"
-            position="inside"
-        >
+        <xpath expr="//table[@name='stock_move_line_table']/thead/tr" position="inside">
             <t t-if="o.valued and o.sale_id and o.move_line_ids and is_outgoing">
-                <t t-if="o.state != 'done'">
-                    <th class="text-end"><strong>Qty Reserved</strong></th>
-                </t>
```
nguyenminhchien: i removed this because the table `stock_move_line_table` was only shown when `o.state == done`.

stock_picking_report_valued/report/stock_picking_report_valued.xml
```diff
@@ -92,18 +90,9 @@
...
-            <t t-if="move_line.picking_id.state != 'done'">
-                    <td class="text-end">
-                        <span t-field="move_line.product_uom_qty" />
-                        <span t-field="move_line.product_uom_id" /></td>
-                </t>
-                <t t-elif="move_line.picking_id.state == 'done'">
-                    <td class="text-center">
-                        <span t-field="move_line.qty_done" />
-                        <span t-field="move_line.product_uom_id" /></td>
-                </t>
```
nguyenminhchien: Because the new columns are added to the table `stock_move_line_table` which shown when o.state = done, so, we don't need these lines of code.

gurneyalex (approved):

gurneyalex:
@nguyenminhchien can you add a note in the README file mentioning that this module includes the features from stock_picking_report_valued_delivery (maybe in a  Changelog section)

nguyenminhchien:
> @nguyenminhchien can you add a note in the README file mentioning that this module includes the features from stock_picking_report_valued_delivery (maybe in a Changelog section)

Added.

PR #302, opened 2024-02-26
[16.0][IMP]stock_report_quantity_by_location: return computed stock reports

PicchiSeba:
Returning the computed reports allows to use them for other purposes.
Like printing a report directly from the `stock.report.quantity.by.location.prepare` form


rousseldenis:
@PicchiSeba I'm wondering if this is really necessary as we have the wizard id, we can retrieve the data. A compute function should not return something (by essence)

PicchiSeba:
> @PicchiSeba I'm wondering if this is really necessary as we have the wizard id, we can retrieve the data. A compute function should not return something (by essence)

@rousseldenis 
That's a good point.  Somehow I didn't think about it.

Thanks for your feedback, closing this PR

PR #306, opened 2024-04-05
[14.0][ADD] stock_quant_history: Generate stock.quant at given pasted date

petrus-v:


stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -0,0 +1,180 @@
...
+            )
+
+        return domain
+
+    @api.model
+    def _ignored_location_usage(self):
+        """If you overwrite or change this
+        list you'll probably want to regenerate all your
+        snapshots"""
+        return [
```
alexandregaldeano, line 95: **question:** is there a reason to provide these values by default? It's looks a bit to specific

petrus-v: I don't think it's specific! Snapshot can be very heavy on the database as we are duplicating all stock quants (maybe snapshot is not perfect wording because of this full duplication :thinking: ). 

I prefer by default save the planet using less storage, if business want to keep details years after years the same information over snapshots quantities in supplier for a given lot will still the same ! If some once really needs it in its business he can overwrite this list from its custom module !

alexandregaldeano (approved):

alexandregaldeano (approved):
Still looking good to me ;)

damdam-s (approved):

hparfr (approved):
LGTM, can you squash in one commit ?

petrus-v:
@hparfr thanks for review, I've squashed commits and rebase the PR !

PR #308, opened 2024-04-19
[16.0][FIX] stock_picking_report_valued: Use original qty done column from …

EmilioPascual:
I don't understand why it is necessary to hide the original qty done column. This makes other modules that inherit the qty done column not work.

@Shide @yajo please review.

@moduon MT-5825

### DIFF
--- a/stock_picking_report_valued/report/stock_picking_report_valued.xml
+++ b/stock_picking_report_valued/report/stock_picking_report_valued.xml
@@ -5,20 +5,20 @@
         <xpath expr="//div[hasclass('page')]" position="before">
             <t t-set="is_outgoing" t-value="o.picking_type_code == 'outgoing'" />
         </xpath>
-        <xpath
-            expr="//table[@t-if=&quot;o.move_line_ids and o.state==&apos;done&apos;&quot;]/thead/tr"
-            position="inside"
-        >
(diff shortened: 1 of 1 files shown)


rafaelbn (approved):
Make sense! It should be done with good inheritance 

👍🏼

stock_picking_report_valued/report/stock_picking_report_valued.xml
```diff
@@ -94,36 +88,18 @@
         inherit_id="stock.stock_report_delivery_has_serial_move_line"
     >
         <xpath expr="//td[@name='move_line_lot_qty_done']" position="after">
```
yajo, line 96: suggestion: if possible, it's better to add a `name` attribute to the `th` above and find the corresponding position here in the tbody using [this trick](https://github.com/orgs/OCA/discussions/111). Otherwise, other modules that extend the same report could put some column in the middle and break the layout.

yajo (approved):

stock_picking_report_valued/report/stock_picking_report_valued.xml
EmilioPascual: Cool trick! I save it for the future. Thanks.
Now I fix with `name` for this case.

yajo (approved):

Gelojr (approved):
LGTM

PR #311, opened 2024-04-24
[17.0] [MIG] stock_picking_report_custom_description

David-Luis-Mora:


Bart-dh:
Hi

I don't know if this was the case for previous versions, but I've found one problem with this module:
When using 2step delivery, with pick & deliver steps, the description_picking doesn't propagate correctly for the picking move, as it has no values["sale_line_id"]. It is only connected with the sale order through it's move_dest_ids.

I temporarily solved it like this, but perhaps there is a better way:
image
image

ioans73:
@David-Luis-Mora can you apply the suggestions of @Bart-dh ?

Bart-dh:
@David-Luis-Mora

juancarlosonate-tecnativa:
Suppressed by https://github.com/OCA/stock-logistics-reporting/pull/363

PR #313, opened 2024-04-26
[16.0][FIX] stock_card_report: date report with timezone

Saran440:
Fixed date result with timezone

Example move stock at 01/01/2024 00:00:00 (with timezone)

Selection_066

User timezone UTC+7
Report will show date without timezone 31/12/2023
Selection_067


hparfr (approved):

rousseldenis:
@Saran440 Should be great to add this case to tests

Saran440:
@rousseldenis Thank you for review. I added this case in test script and change `setUp` to `setUpClass`

rousseldenis (approved):

PR #336, opened 2024-08-28
[16.0][IMP] stock_average_daily_sale: support multi-warehouses

twalter-c2c:
1. Dependencies to "abc profile" has been removed as consumptions are per warehouse, independently of the many profile you could have on a warehouse.
2. Major changes in the configuration: added company field, added form view, allowed excluding or not the weekends for calculating values of the report.
3. Changes in the query (which calculates report values) logic: exclude weekends only if this option is chosen in the config, include production location as destination location.


simahawk:
@jbaudoux ciaoooooooo ping

stock_average_daily_sale/__manifest__.py
```diff
@@ -5,24 +5,20 @@
     "name": "Stock Average Daily Sale",
     "summary": """
         Allows to gather delivered products average on daily basis""",
-    "version": "16.0.1.0.0",
+    "version": "16.0.2.0.0",
```
jbaudoux: Don't change the version yourself unless there is a migration script, it will be done automatically by the ocabot

stock_average_daily_sale/data/ir_cron.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo noupdate="1">
+
```
jbaudoux: Can you reduce the diff by removing all changes that are not necessary? This includes the renaming of "as" into "AS" in the query, added or removed blank lines (unless pre-commit complains)

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -58,12 +52,6 @@ class StockAverageDailySale(models.Model):
     )
     date_from = fields.Date(string="From", required=True)
     date_to = fields.Date(string="To", required=True)
-    is_mto = fields.Boolean(
-        string="On Order",
-        readonly=True,
-        store=True,
-        index=True,
-    )
```
jbaudoux: Why do you remove this?

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -9,28 +9,37 @@
...
+    check_company_auto = True
 
-    abc_classification_profile_id = fields.Many2one(
-        comodel_name="abc.classification.profile",
-        required=True,
-        ondelete="cascade",
-    )
     abc_classification_level = fields.Selection(
-        selection=ABC_SELECTION, required=True, readonly=True
+        selection=ABC_SELECTION, required=True, default="b"
```
jbaudoux, line 17: Either we always need a config for each a/b/c and it's required & readonly & auto-generated for each warehouse, or you make it optional & non readonly. The second option would be an improvement that is not part of your list of improvements. I would not change that for now.

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -9,28 +9,37 @@
...
     )
-    standard_deviation_exclude_factor = fields.Float(required=True, digits=(2, 2))
-    warehouse_id = fields.Many2one(
+    name = fields.Char(string="Name", required=True)
```
jbaudoux: No need for this new field

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -9,28 +9,37 @@
...
+    company_id = fields.Many2one(
+        string="Company",
+        comodel_name="res.company",
+        required=True,
+        default=lambda self: self.env.company,
+    )
```
jbaudoux: You need to use the company of the warehouse

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -9,28 +9,37 @@
...
+    standard_deviation_exclude_factor = fields.Float(
+        required=True, default=0, digits=(2, 2)
+    )  # TODO check if 0 is correct as default
+    warehouse_ids = fields.Many2many(
```
jbaudoux: The consumption is computed per product per warehouse (see deliveries_last), so having one config per warehouse is legit. It looks to me like there is a missing where clause in `deliveries_last` to take the config of the warehouse

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -9,28 +9,37 @@
...
-    warehouse_id = fields.Many2one(
+    name = fields.Char(string="Name", required=True)
+    company_id = fields.Many2one(
+        string="Company",
+        comodel_name="res.company",
+        required=True,
+        default=lambda self: self.env.company,
+    )
+    standard_deviation_exclude_factor = fields.Float(
+        required=True, default=0, digits=(2, 2)
```
jbaudoux: It already defaults to 0, you don't need to add such default

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -46,4 +55,4 @@ class StockAverageDailySaleConfig(models.Model):
     number_days_qty_in_stock = fields.Integer(
         string="Number of days of quantities in stock", required=True, default=2
     )
-    safety_factor = fields.Float(digits=(2, 2), required=True)
+    safety_factor = fields.Float(digits=(2, 2), required=True, default=0)
```
jbaudoux: It already defaults to 0, you don't need to add such default

stock_average_daily_sale/static/description/index.html
```diff
@@ -8,10 +8,11 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
```
jbaudoux: Exclude from your changes commit the generated files
You can use `export SKIP=oca-gen-addon-readme`

stock_average_daily_sale/views/stock_average_daily_sale.xml
```diff
@@ -16,16 +16,6 @@
...
-                <filter
-                    string="Normal product"
-                    name="normal_product"
-                    domain="[('is_mto','=',False)]"
-                />
-                <filter
-                    string="On Order"
-                    name="on_order"
-                    domain="[('is_mto','=',True)]"
-                />
```
jbaudoux: Keep this

jbaudoux (changes requested):
Some remarks

jbaudoux:
Can you also prefix your commits first line message by the module name `stock_average_daily_sale: ...`

jbaudoux:
cc @rousseldenis

stock_average_daily_sale/static/description/index.html
twalter-c2c: Thank you for this tip. I excluded the file.

stock_average_daily_sale/models/stock_average_daily_sale_config.py
twalter-c2c: Yes, probably by default one config per warehouse is fine. I would need to allow editing multiple configs at once, but I will do it in a custom module.

stock_average_daily_sale/models/stock_average_daily_sale_config.py
twalter-c2c: @jbaudoux Is such config auto-generated now? Could you please give me a hint how? If it is the case I probably can leave it unchanged indeed. However, I only noticed the "auto-config" in a demo mode. So I changed it to allow users to modify the config easily.

jbaudoux: You're right, your change is legit then. Could add a unique sql constraint on warehouse and abc_classification_level ?

jbaudoux (commented):
Looks good. Thanks.

* Can you reword your commits to prefix first line message by the module name? stock_average_daily_sale: ...
* Can you add the unique constraint on the config?

twalter-c2c:
Thank you for the review. Yes, I was just about to add the prefix to commit messages, forgot about it while pushing the last changes. I will also add a constraint then, thank you for the suggestion.

stock_average_daily_sale/views/stock_warehouse.xml
```diff
@@ -10,7 +10,11 @@
         <field name="arch" type="xml">
             <xpath expr="//group[1]" position="after">
                 <group string="Average Sales" name="average_sale_reporting">
-                    <field name="average_daily_sale_root_location_id" />
+                    <field name="lot_stock_id" invisible="1" />
```
twalter-c2c, line 13: I needed to add this field to the view because of the error
```Field 'lot_stock_id' used in attrs ({'required': [('lot_stock_id', '!=', False)]}) is restricted to the group(s) stock.group_adv_location,stock.group_stock_multi_warehouses,base.group_no_one.```
I've got this error after adding the required attribute to the field below. Adding the abovementioned groups to the `average_daily_sale_root_location_id` field or to it's parent view group doesn't solve the issue while adding invisible `lot_stock_id` field does.

File: stock_average_daily_sale/models/stock_average_daily_sale_config.py [hunk 13]
```diff
@@ -47,3 +52,11 @@ class StockAverageDailySaleConfig(models.Model):
         string="Number of days of quantities in stock", required=True, default=2
     )
     safety_factor = fields.Float(digits=(2, 2), required=True)
+
+    _sql_constraints = [
+        (
+            "abc_classification_level_unique",
+            "UNIQUE(abc_classification_level)",
+            _("Abc Classification Level must be unique"),
```
jbaudoux:
```suggestion
            "UNIQUE(abc_classification_level, warehouse_id)",
            _("Abc Classification Level must be unique per warehouse"),
```

stock_average_daily_sale/views/stock_warehouse.xml
jbaudoux: Yes, I saw it and what you did is the correct way of doing :)

jbaudoux (approved):

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -30,7 +31,11 @@ class StockAverageDailySaleConfig(models.Model):
         default=lambda self: self.env["stock.warehouse"].search(
             [("company_id", "=", self.env.company.id)], limit=1
         ),
-        readonly=True,
+    )
+    exclude_weekends = fields.Boolean(
+        help="Set to True only if you do not expect any orders/deliveries during "
+        "the weekends. If set to True, stock moves done on weekends won't be "
+        "taken into account to calculate the average daily usage",
     )
```
jbaudoux, line 39: I forgot you need a pre-migration script to add this new field and set it to True as that was the previous behavior.

jbaudoux (changes requested):

stock_average_daily_sale/models/stock_average_daily_sale_config.py
twalter-c2c: Yes, that is true, I will add it.

stock_average_daily_sale/migrations/16.0.1.1.0/pre-migrate.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2024 Camptocamp SA (http://www.camptocamp.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```
jbaudoux: To be consistent with manifest
```suggestion
# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```

jbaudoux (approved):

henrybackman (approved):
Technical review only

simahawk:
@rousseldenis good for you?

rousseldenis (approved):

stock_average_daily_sale/tests/common.py
```diff
@@ -87,12 +87,13 @@ def _create_move(cls, product, origin_location, qty):
                 "product_id": product.id,
                 "name": product.name,
                 "location_id": origin_location.id,
+                "warehouse_id": origin_location.warehouse_id.id,
```
jbaudoux, line 90: The warehouse must come from the picking_type_id on the move

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -228,9 +225,10 @@ def _create_materialized_view(self):
                         JOIN cfg on cfg.abc_classification_level = coalesce(pt.abc_storage, 'c')
                     WHERE
                         sl_src.usage in ('view', 'internal')
-                        AND sl_dest.usage = 'customer'
+                        AND sl_dest.usage in ('customer', 'production')
                         AND sm.date BETWEEN cfg.date_from AND cfg.date_to
                         AND sm.state = 'done'
+                        AND sm.warehouse_id = cfg.warehouse_id
```
jbaudoux, line 231: warehouse must come from the sm.picking_type_id

jbaudoux (commented):
The multi-warehouse needs fixing

PR #337, opened 2024-09-04
[16.0][ADD] stock_account_valuation_discrepancy_adjust

AaronHForgeFlow:
Trying to add a module to fix discrepancies in the Dual inventory valuation report. The first attempt was this one: https://github.com/OCA/stock-logistics-reporting/pull/175

This version is much simpler because the customer don't want to have a separate menu, I think it is not needed. The idea is they run the current report, and then the user uses the filter "Valuation Discrepancy" in order to see the products with discrepancies and apply the adjustments.

If users wants to do the adjustment in the past then they have to generate the Dual Inventory Valuation report in the past.

Added a small test also

Added the option to create a single journal entry  if wanted.

cc @ForgeFlow


AaronHForgeFlow:
In order to do a functional review you can do the following:

Create a storable product with standard cost .Ensure the product category is real time valuated:


Update the stock on hand:



Update the stock on hand again, and apply the changes:




Unpost a journal entry created by one of the stock changes (search for "Product Quantity Updated"):



Run the Dual Inventory Valuation report (in Inventory / Reporting / Dual Inventory Valuation) (Please notice the report is already in OCA, here I am adding the wizard only). Open the wizard for the product:



Choose the counterpart accounts and click on "Make Adjustment":



Run the report "Dual Inventory Valuation" again. The product should show no discrepancy now :)

You can check the journal entry created by looking at the journal entry with reference "Adjust for Stock Valuation Discrepancy"

LauraCForgeFlow (approved):
LGTM!

MiquelRForgeFlow (approved):
LGTM

stock_account_valuation_discrepancy_adjust/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "license": "AGPL-3",
+    "depends": ["stock_account_valuation_report"],
+    "data": [
+        "security/ir.model.access.csv",
+        "wizards/wizard_stock_discrepancy_adjustment_view.xml",
+    ],
+    "demo": [],
+    "installable": True,
+    "development_status": "Alpha",
+    "maintainers": ["ChrisOForgeFlow"],
```
LoisRForgeFlow: I think you will need to adopt this one

AaronHForgeFlow: Good point! I did :)

PR #342, opened 2024-09-24
[MIG] stock_average_daily_sale: Backporting to 14.0 from 16.0

twalter-c2c:


simahawk:
@twalter-c2c we have a maturity level conflict
image

See https://github.com/OCA/stock-logistics-warehouse/blob/14.0/stock_location_warehouse/__manifest__.py#L14
It can be raised to Beta if not Stable as is there since ages.

stock_average_daily_sale/models/stock_warehouse.py
```diff
@@ -29,11 +28,11 @@ def _compute_average_daily_sale_root_location_id(self):
                 continue
             warehouse.average_daily_sale_root_location_id = warehouse.lot_stock_id
 
-    @api.model_create_multi
```
simahawk: why changing this?

stock_average_daily_sale/tests/test_average_daily_sale.py
```diff
@@ -188,10 +190,20 @@ def test_average_sale_profile_a(self):
 
     def test_view_refreshed(self):
         self._refresh()
-        with self.assertNoLogs(
+        with self.assertLogs(
             "odoo.addons.stock_average_daily_sale.models.stock_average_daily_sale",
             level="DEBUG",
-        ):
+        ) as cm:
+            logging.getLogger(
```
simahawk, line 201: I'm not sure to understand this

simahawk (changes requested):

stock_average_daily_sale/migrations/16.0.1.1.0/pre-migrate.py
henrybackman, line 1: Not sure if this is needed when backporting to 14 or should it be renamed to be `14.0.1.1.0` instead of the `16.0.1.1.0`

twalter-c2c: You are right, the version is not correct, but anyway I think it's not needed when backporting. I removed the `migrations` from the module.

stock_average_daily_sale/models/stock_warehouse.py
twalter-c2c: As discussed at the meeting: it reflects differences between 16.0 and 14.0 in the Odoo standard stock_warehouse create override.

stock_average_daily_sale/tests/test_average_daily_sale.py
twalter-c2c: I added a comment explaining why I changed it and why in that way.

simahawk (approved):

henrybackman (approved):

simahawk:
@rousseldenis would you mind to have a look? :pray:

stock_average_daily_sale/tests/common.py
```diff
@@ -0,0 +1,119 @@
...
+                "usage": "inventory",
+            }
+        )
+        cls.stock_location = cls.env.ref("stock.warehouse0").lot_stock_id
+
+        cls._create_products()
+
+    @classmethod
+    def _create_inventory(cls):
+        inventory = cls.env["stock.inventory"].create(
```
sebalix, line 53:
```suggestion
        inventory = cls.inventory_obj.create(
```

stock_average_daily_sale/tests/common.py
```diff
@@ -50,20 +50,34 @@ def setUpClass(cls):
 
     @classmethod
     def _create_inventory(cls):
-        cls.inventory_obj.create(
+        inventory = cls.env["stock.inventory"].create(
```
sebalix, line 53: Not sure why we switched from `stock.quant` to `stock.inventory` to put goods in stock, we can also use quant model to do that in v14. E.g:
```python
        cls.env["stock.quant"]._update_available_quantity(
            product,
            location,
            quantity,
            package_id=package,
            lot_id=lot,
            in_date=in_date,
        )
```
Or if we want to state the current qty in stock (not adding new ones), we can use this helper method:
```python
    @classmethod
    def _update_qty_in_location(
        cls, location, product, quantity, package=None, lot=None, in_date=None
    ):
        quants = cls.env["stock.quant"]._gather(
            product, location, lot_id=lot, package_id=package, strict=True
        )
        # this method adds the quantity to the current quantity, so remove it
        quantity -= sum(quants.mapped("quantity"))
        cls.env["stock.quant"]._update_available_quantity(
            product,
            location,
            quantity,
            package_id=package,
            lot_id=lot,
            in_date=in_date,
        )
```

But probably not blocking for a backport.

sebalix (approved):

sebalix:
I found an issue, if there is no `warehouse_id` defined on stock moves (it's not required), the sql request doesn't provide any result. Fix ongoing.

EDIT: fixed

twalter-c2c:
FTR: I noticed that previous day's moves were not being included in the calculations (although they were supposed to be). The last commit I posted a week ago fixes that issue.

jbaudoux (approved):

jbaudoux:
@twalter-c2c Can you port the 2 fixes to 16.0 ?

twalter-c2c:
> @twalter-c2c Can you port the 2 fixes to 16.0 ?

@jbaudoux Of course, I will do that next week. Thank you for merging the PR.

PR #345, opened 2024-10-04
[16.0][IMP]stock_report_quantity_by_location: Included report generation

anusriNPS:
Included pdf report generation functionality for module  "stock_report_quantity_by_location"
    
    Reports uses different html tags for view and print report functionality:
            - View report uses div tags with classes defined in css file
            - Printing report uses table related tags such as thead, tr, td etc based on web.internal layout used in template


anusriNPS:
View report and export pdf report:

Screenshot from 2024-10-04 13-25-09
Screenshot from 2024-10-04 13-24-12

stock_report_quantity_by_location/wizards/stock_report_quantity_by_location_prepare.py
```diff
@@ -79,3 +84,40 @@ def _compute_stock_report_by_location(self):
...
+            "action_stock_report_quantity_by_location_html"
+        )
+        vals = action.read()[0]
+        new_context = vals.get("context", {})
+        if isinstance(new_context, str):
+            try:
+                new_context = safe_eval(new_context)
+            except (TypeError, SyntaxError, NameError, ValueError):
+                _logger.warning(
+                    _("Failed context evaluation: %(context)s", context=new_context)
```
PicchiSeba: **suggestion**: let's maintain the same string translation/formatting structure as the other modules

```suggestion
                    _("Failed context evaluation: %(context)s") % {"context": new_context}
```

stock_report_quantity_by_location/wizards/stock_report_quantity_by_location_prepare.py
```diff
@@ -79,3 +84,40 @@ def _compute_stock_report_by_location(self):
...
                     )
         self.env["stock.report.quantity.by.location"].create(vals_list)
+
+    def button_export_html(self):
+        self.ensure_one()
+        action = self.env.ref(
+            "stock_report_quantity_by_location."
+            "action_stock_report_quantity_by_location_html"
+        )
+        vals = action.read()[0]
```
PicchiSeba: **issue**: we should use the proper way to open an action

```suggestion
        action = self.env["ir.actions.act_window"]._for_xml_id(
            "stock_report_quantity_by_location."
            "action_stock_report_quantity_by_location_html"
        )
```

Following this change, rename the following instances of `vals` with `action`

PicchiSeba (changes requested):
Code review: minor fixes

stock_report_quantity_by_location/wizards/stock_report_quantity_by_location_prepare.py
anusriNPS: Modified as suggested

stock_report_quantity_by_location/wizards/stock_report_quantity_by_location_prepare.py
anusriNPS: Used "ir.actions.actions" to get an action

PicchiSeba (approved):
Code review: LGTM

PR #351, opened 2024-11-10
[16.0][IMP] stock_quantity_history_location: add usage view

lfreeke:
Let the user select usage (location type) 'view' to select all child location. 
This is useful if you have a location that has a direct parent to the usage (location type) 'view'.


rousseldenis:
Hi @lfreeke thanks for this. Seems legit.

Could you check why tests are failing ?

lfreeke:
> Hi @lfreeke thanks for this. Seems legit.
> 
> Could you check why tests are failing ?

I retriggerd the build and now everything is green

PR #353, opened 2024-12-03
[16.0][FIX] stock_picking_report_custom_description: avoid dupe product name

chienandalu:
By default the sale description is set as
f"{product_name}\n{description". If we pass that string as the description_picking, when we print the report, the name will be duplicated. To avoid that effect, we want to trim the product name from the description passed to the picking by default.

cc @Tecnativa TT52044

please review @pedrobaeza @pilarvargas-tecnativa

### DIFF
--- a/stock_picking_report_custom_description/models/stock_rule.py
+++ b/stock_picking_report_custom_description/models/stock_rule.py
@@ -31,6 +31,15 @@ class StockRule(models.Model):
         )
         if values.get("sale_line_id"):
             line = self.env["sale.order.line"].browse(values["sale_line_id"])
-            res["description_picking"] = line.name
+            # Avoid double printing the name in the picking
+            pattern = f"{line.product_id.display_name}\n"
+            description_picking = line.name
+            if description_picking.startswith(pattern):
+                description_picking = description_picking.replace(pattern, "")
+            if (
+                description_picking
+                and description_picking != line.product_id.display_name
+            ):
+                res["description_picking"] = description_picking
             res["name"] = line.name
         return res
--- a/stock_picking_report_custom_description/tests/test_stock_picking_report_custom_description.py
+++ b/stock_picking_report_custom_description/tests/test_stock_picking_report_custom_description.py
@@ -11,13 +11,16 @@ class TestStockPickingReportCustomDescription(common.TransactionCase):
         super().setUpClass()
         cls.customer = cls.env["res.partner"].create({"name": "Test customer"})
         cls.product = cls.env["product.product"].create(
-            {"name": "Test product", "type": "product"}
+            {
+                "name": "Test product",
+                "type": "product",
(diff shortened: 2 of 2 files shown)


stock_picking_report_custom_description/models/stock_rule.py
```diff
@@ -31,6 +31,14 @@ def _get_stock_move_values(
         )
         if values.get("sale_line_id"):
             line = self.env["sale.order.line"].browse(values["sale_line_id"])
-            res["description_picking"] = line.name
+            # Avoid printing double printing the name in the picking
```
pedrobaeza:
```suggestion
            # Avoid double printing the name in the picking
```

stock_picking_report_custom_description/models/stock_rule.py
```diff
@@ -31,6 +31,14 @@ def _get_stock_move_values(
...
+            description_picking = line.name.replace(
```
pedrobaeza: Replacing unconditionally whole `display_name` seems too much, isn't it? Shouldn't it be only removed in case of having something more?

```python
if line != line.product_id.display_name:
```

chienandalu: It's what I do in the line below. And I only replace it in case of matching the pattern `f"{product.display_name}\n", not just the display name. That pattern comes from the method `get_product_multiline_description_sale`. It basically goes like this:

```python
name = self.display_name
if self.description_sale:
    name += '\n' + self.description_sale
```

So I just want to look after that issue.

pedrobaeza: Well, not exactly, as you are doing an unconditional replace in any part of the text. Imagine a product called `Test` and this SOL line text:

```
Test
This is a Test
Another line
```

you will end up with

```
This is a Another Line
```

chienandalu: Ok, I can improve it with a regex match :sweat_smile:

pedrobaeza: Or simply an `startswith` plus a replace with just 1 repetition.

pilarvargas-tecnativa (approved):

stock_picking_report_custom_description/models/stock_rule.py
pedrobaeza: This one is pending.

stock_picking_report_custom_description/models/stock_rule.py
pedrobaeza: I don't know if my suggestion is faster, but usually everything is faster than regex.

chienandalu:
@pedrobaeza Changes done

pedrobaeza (approved):
/ocabot merge patch

PR #354, opened 2024-12-06
[14.0][ADD] auto_printing_stock_picking

jbaudoux:
When a picking is done, automatically trigger the printing of some documents.
This can be used to print a delivery slip (report) or labels received from the carrier (attachment).

Depends on:
* https://github.com/OCA/report-print-send/pull/310


jbaudoux:
cc @mt-software-de @sebalix

mt-software-de (approved):
LGTM

simahawk (commented):
LG

printing_auto_stock_picking/models/stock_picking_type.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    auto_printing_ids = fields.Many2many(
+        "printing.auto",
+        string="Auto Printing Configuration",
+        domain=[("model", "=", "stock.picking.type")],
```
jbaudoux: @mt-software-de I wonder if this is not a mistake and if we should not put the model of what is printed instead of where it is configured? And in `_get_printing_auto` filter on that model. https://github.com/OCA/report-print-send/pull/310/files#diff-1360f4baf5029947c074c2576244e65952f4043906416d240f5026996ac26f68R37

mt-software-de: Maybe I don't quite understand what you mean.
We need to configure the printing.auto on the picking.type.
I also think we should change the model in the domain to stock.picking. This then aligned with the model which is printing. 
But a filter is not needed, because the auto_printing_ids are on the model with the mixin.

printing_auto_stock_picking/views/stock_picking_type.xml
```diff
@@ -0,0 +1,16 @@
...
+  <record id="view_picking_type_form_inherit" model="ir.ui.view">
+   <field name="name">printing.auto.stock.picking</field>
+    <field name="model">stock.picking.type</field>
+    <field name="inherit_id" ref="stock.view_picking_type_form" />
+    <field name="arch" type="xml">
+      <xpath expr="//sheet/group[last()]">
+        <separator string="Auto Printing Configuration" colspan="4" />
+        <field
+                    name="auto_printing_ids"
+                    context="{'default_model': 'stock.picking.type'}"
```
mt-software-de: Could you also Change The model here?
```suggestion
                    context="{'default_model': 'stock.picking'}"
```

printing_auto_stock_picking/models/stock_picking_type.py
mt-software-de: Yes we should change the Domain to `stock.picking` this makes more sense. And change it on the view and it should be fine

mt-software-de (changes requested):

printing_auto_stock_picking/models/stock_picking_type.py
jbaudoux: fixed in https://github.com/OCA/stock-logistics-reporting/pull/354/commits/c137eb423f6daaf07c928cacd13f2eac79f2c9f8

jbaudoux:
@mt-software-de I fixed the model and added a migration script

mt-software-de (approved):
LGTM

PR #357, opened 2024-12-10
[14.0][FIX] stock_average_daily_sale: make sure joining on unique ids

twalter-c2c:
Concatenation of `warehouse_id` and `product_id` may have resulted in duplicated values in case of having more than 10 warehouses (`select concat(11, 2)::integer = concat(1, 12)::integer` -> true) impacting the `join` in materialized view query (`JOIN daily_standard_deviation ds on ds.id= t.window_id`). With this PR we make sure to have a unique value after concatenation.


sebalix (approved):
LG, this will be forward-ported to 16.0 once merged

jbaudoux (approved):

PR #358, opened 2024-12-12
[15.0][IMP] stock_picking_report_custom_description: update description

chienandalu:
Partially backported from the v16 feature

Before this change only when the procurements where updated the sale line description would be propagated so if a user updated that description after the sale was confirmed it wouldn't be written to the picking move.

Now every time the user changes the line it will be synced to the related moves.

cc @Tecnativa TT52171

please review @pilarvargas-tecnativa @carlos-lopez-tecnativa

### DIFF
--- a/stock_picking_report_custom_description/models/sale_order_line.py
+++ b/stock_picking_report_custom_description/models/sale_order_line.py
@@ -1,0 +1,16 @@
+# Copyright 2023 Tecnativa - Carolina Fernandez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("name"):
+            self.env["stock.move"].search(
+                [("sale_line_id", "in", self.ids)]
+            ).name = vals["name"]
+        return res
--- a/stock_picking_report_custom_description/README.rst
+++ b/stock_picking_report_custom_description/README.rst
@@ -30,6 +30,7 @@ Stock Picking Report Custom Description
 
 This module transfers the sales order line description to the picking, and
 allows to print such text on the picking reports.
+When the sales order line description is changed, picking description is updated.
 
 **Table of contents**
 
--- a/stock_picking_report_custom_description/models/__init__.py
+++ b/stock_picking_report_custom_description/models/__init__.py
@@ -1,2 +1,3 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from . import stock_rule
+from . import sale_order_line
--- a/stock_picking_report_custom_description/readme/DESCRIPTION.rst
+++ b/stock_picking_report_custom_description/readme/DESCRIPTION.rst
@@ -1,2 +1,3 @@
 This module transfers the sales order line description to the picking, and
 allows to print such text on the picking reports.
+When the sales order line description is changed, picking description is updated.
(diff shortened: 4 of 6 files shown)


stock_picking_report_custom_description/models/sale_order_line.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("name"):
+            self.env["stock.move"].search([("sale_line_id", "=", self.id)]).name = vals[
```
rousseldenis: This is not correct.

```suggestion
            self.env["stock.move"].search([("sale_line_id", "in", self.ids)]).update({"name": vals["name"]})
```

rousseldenis (changes requested):

stock_picking_report_custom_description/models/sale_order_line.py
pedrobaeza: OK about the `in` operator, but you don't need `update`, as dot notation serves also for writing in a multi-record.

chienandalu: Thanks @rousseldenis :). I just changed. As @pedrobaeza said, we don't need the `update` anyway. I'll check v16, as this is a potential bug.

stock_picking_report_custom_description/models/sale_order_line.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("name"):
+            self.env["stock.move"].search(
+                [("sale_line_id", "in", self.ids)]
```
pedrobaeza, line 14: I'm now thinking that this should only be applied for pending pickings, not already done ones.

stock_picking_report_custom_description/models/sale_order_line.py
```diff
@@ -0,0 +1,16 @@
...
+            ).name = vals["name"]
```
carlos-lopez-tecnativa, line 15: Theoretically, this can raise a singleton error. Additionally, why not use the `move_ids` field instead of performing a search?
```suggestion
            self.env["stock.move"].search([("sale_line_id", "in", self.ids)]).name = vals[
                "name"
            ]
```

stock_picking_report_custom_description/models/sale_order_line.py
pedrobaeza: @carlosdauden what's your take on this? Do you think it's OK to change done pickings?

stock_picking_report_custom_description/models/sale_order_line.py
carlos-lopez-tecnativa: Sorry, I see that ids was updated while I was writing my review.

stock_picking_report_custom_description/models/sale_order_line.py
rousseldenis: Yes of course

stock_picking_report_custom_description/models/sale_order_line.py
carlosdauden: If the data is changed to improve it, we will certainly want it to be modified everywhere.
If the data is changed to make it worse, we will not want it to be modified in the delivery notes that have already been made (although we would not really want it to change anywhere).

But well, as any modification must be assumed to be made to improve, I prefer that all the pickings are updated.
So that in case the user makes a mistake, he can always rectify it.

carlosdauden (approved):

stock_picking_report_custom_description/models/sale_order_line.py
pedrobaeza: OK then. Let's keep it.

pedrobaeza (approved):
@rousseldenis please remove your "requested changes".

rousseldenis (approved):

PR #359, opened 2024-12-12
[16.0][FIX] stock_picking_report_custom_description: singleton error

chienandalu:
As reported in

- #358 

It isn't very common to write the name in batch. Anyway it would lead to errors in it's current shape.

cc @Tecnativa TT52171

please review @carlos-lopez-tecnativa @pedrobaeza


carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #364, opened 2025-01-13
[16.0][FIX] stock_account_valuation_discrepancy_adjust: Let the user change the date of the adjustment

AaronHForgeFlow:
Adjustment on past custom date should be possible:

image

Currently all adjustments are created only on today's date.

cc @ForgeFlow

### DIFF
--- a/stock_account_valuation_discrepancy_adjust/tests/test_stock_account_valuation_discrepancy_adjust.py
+++ b/stock_account_valuation_discrepancy_adjust/tests/test_stock_account_valuation_discrepancy_adjust.py
@@ -3,6 +3,8 @@
 
 import time
 
+from dateutil.relativedelta import relativedelta
+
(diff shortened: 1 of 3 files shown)


gdgellatly:
@AaronHForgeFlow are we sure this is the right place to do this. Is it not better to ensure the date from the wizard action is passed. The reason I ask, is that the selection lines are passed from the report, to me it doesn't make sense that we would select some products at one date in the report, then set a different date where the discrepancies may be entirely different from the report.

EDIT: I found that this is already the behaviour, however there is a bug in the move date. I fixed in my PR, however currently the to_date is a datetime and in the journal entry it isn't made context aware so can be wrong date

AaronHForgeFlow:
Closing this in favor of https://github.com/OCA/stock-logistics-reporting/pull/377

PR #367, opened 2025-02-03
[18.0][MIG] stock_picking_report_undelivered_product: Migration to 18.0

juancarlosonate-tecnativa:
The module now renders its own table instead of reusing the standard backorder block, which can be replaced by other modules (e.g. stock_picking_group_by_partner_by_carrier); it keeps the same title and columns, so look and stay consistent. The filters were adapted to the new quantity semantics (on non-done moves it holds the reserved quantity): the module lists the quantities cancelled when validating without backorder, leaving pending backorder quantities to the standard block, as in 15.0. The customer checkbox was also moved to an always visible group (the stock partner "Inventory" group requires developer mode).

@Tecnativa TT54343

@CarlosRoca13 @sergio-teruel can you check? thanks!


stock_picking_report_undelivered_product/views/report_deliveryslip.xml
```diff
@@ -1,95 +1,56 @@
...
-            <!-- Get stock moves from backorder according to setting values -->
-            <t
-                t-if="not o.company_id.undelivered_product_slip_report_method or o.company_id.undelivered_product_slip_report_method == 'all'"
-            >
+    <template
+        id="undeliverd_product"
+        inherit_id="stock.report_delivery_document"
+        priority="200"
+    >
+        <xpath expr="//*[@t-set='backorders']" position="replace">
```
sergio-teruel: Upps! Try not replace any view!! Hide it instead!

sergio-teruel (changes requested):
Great @juancarlosonate-tecnativa  !!!

Minor fix!!

sergio-teruel:
Review tests!!

rousseldenis (changes requested):
Agree with @sergio-teruel. Don't replace views

stock_picking_report_undelivered_product/views/report_deliveryslip.xml
juancarlosonate-tecnativa: Done!

CarlosRoca13 (approved):
Thanks!

pedrobaeza:
@rousseldenis

juancarlosonate-tecnativa:
@rousseldenis @sergio-teruel check again please!

stock_picking_report_undelivered_product/views/report_deliveryslip.xml
```diff
@@ -0,0 +1,63 @@
...
+        <xpath
+            expr="//span[@t-field='bo_line.product_uom_qty']/.."
+            position="attributes"
+        >
+            <attribute name="class" add="d-none" separator=" " />
+        </xpath>
+
+        <xpath expr="//span[@t-field='bo_line.product_uom_qty']/.." position="after">
+            <td class="text-center w-auto">
+                <span t-esc="bo_line.product_uom_qty - bo_line.quantity">3.00</span>
```
sergio-teruel: Why 3.0?

sergio-teruel: Upsss I see... it's in original view...
Do you know that you can user attribute 'move' to move xml elements??

sergio-teruel: You can see here https://github.com/odoo/odoo/blob/7de8f1edaa91a987161774d6b86220f58ea3354d/addons/project/views/rating_rating_views.xml#L33

juancarlosonate-tecnativa: i know it exists, @CarlosRoca13  mentioned it to me, but i don't know how to use it. Can you give me some more details?

juancarlosonate-tecnativa: please check again

stock_picking_report_undelivered_product/models/stock_move.py
```diff
@@ -1,20 +0,0 @@
-# Copyright 2020 Tecnativa - Sergio Teruel
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
-from odoo import fields, models
-
-
-class StockMove(models.Model):
-    _inherit = "stock.move"
-
-    splitted_stock_move_orig_id = fields.Many2one(
```
rousseldenis: @juancarlosonate-tecnativa Do we lose the information on existing databases ?

juancarlosonate-tecnativa: this information is no longer useful and will never be used

sergio-teruel: Maybe this change needs to be analysed in more depth.

pedrobaeza:
@carlosdauden @juancarlosonate-tecnativa can you check this?

RobbeDynapps:
@carlosdauden @juancarlosonate-tecnativa Can you check again?

carlosdauden:
Can you rebase to force runboat build?

juancarlosonate-tecnativa:
> Can you rebase to force runboat build?

Done

carlosdauden:
I'm testing on Runboat with a delivery order that contains two products: one is partially delivered, and the other is not delivered at all. I validate without creating a backorder, and when I print, it doesn't matter which option I configure in the settings — it always shows all the quantities still to be delivered.
http://oca-stock-logistics-reporting-18-0-pr367-7cf7683e2915.runboat.odoo-community.org/odoo/sales/25/action-365/24?debug=1

juancarlosonate-tecnativa:
> I'm testing on Runboat with a delivery order that contains two products: one is partially delivered, and the other is not delivered at all. I validate without creating a backorder, and when I print, it doesn't matter which option I configure in the settings — it always shows all the quantities still to be delivered. http://oca-stock-logistics-reporting-18-0-pr367-7cf7683e2915.runboat.odoo-community.org/odoo/sales/25/action-365/24?debug=1

@carlosdauden Yes, and in principle it’s correct. Even if you don’t create a backorder, the undelivered quantities are still displayed. That’s how it worked in previous versions.

carlosdauden:
When the ‘completely_undelivered’ option is selected in the ‘undelivered_product_slip_report_method’ field, the behaviour should change and the partially delivered movement should not be displayed.

@sergio-teruel, does this not work this way in 15.0?

juancarlosonate-tecnativa:
Some changes were needed to make the migration work properly. Please review again, thanks!

PR #368, opened 2025-02-05
[MIG] stock_account_valuation_report: Migration to 18.0

StefanRijnhart:
Includes https://github.com/OCA/stock-logistics-reporting/pull/378


stock_account_valuation_report/wizards/stock_valuation_history.py
```diff
@@ -0,0 +1,36 @@
...
+# Copyright 2019 Aleph Objects, Inc.
+from odoo import fields, models
+from odoo.osv import expression
+from odoo.tools.safe_eval import safe_eval
+
+
+class StockValuationHistory(models.TransientModel):
+    _name = "stock.valuation.history"
+    _description = "Stock Valuation History"
+
```
gdgellatly, line 12:
```suggestion
_rec_name = "inventory_datetime"

```

gdgellatly: I made same PR for 16.0 at #378

stock_account_valuation_report/models/product_product.py
```diff
@@ -30,29 +30,11 @@ class ProductProduct(models.Model):
         compute="_compute_inventory_value",
         search="_search_qty_discrepancy",
     )
-    valuation = fields.Selection(
```
gdgellatly: Reviewed and understood reason for deletion. Field is now related on product.template to categ_id and directly accessible to product.

stock_account_valuation_report/wizards/stock_valuation_history.py
StefanRijnhart: PIcked from #378, thanks.

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,191 @@
...
+            != 0
+        )
+        return [("id", "in", products_with_discrepancy.ids)]
+
+    @api.model
+    def _search_valuation_discrepancy(self, operator, value):
+        products = self.with_context(active_test=False).search(
+            [("is_storable", "=", True)],
+        )
+        dp = self.env.ref("product.decimal_discount").precision_get("Discount")
```
gdgellatly: Hmmmm, this looks wrong, I think `self.env.company.currency_id.rounding` and then precision_rounding kw arg in float compare

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,191 @@
...
+        compute="_compute_inventory_value",
+        search="_search_qty_discrepancy",
+    )
+
+    @api.model
+    def _search_qty_discrepancy(self, operator, value):
+        products = self.with_context(active_test=False).search(
+            [("is_storable", "=", True)],
+        )
+        dp = self.env["decimal.precision"].precision_get("Product Price")
```
gdgellatly: This looks wrong, price != quantity, I think it should be UoM precision, I made similar changes in my recent PR for discrepancy adjustment, use UoM digits here

gdgellatly (changes requested):
I actually found these in the discrepancy module today, maybe we need it here too https://github.com/OCA/stock-logistics-reporting/pull/377/files#diff-d6055728635c3de730f384efe25ffbea9ac93cf0417f7337efb3b2a7fb88510e

stock_account_valuation_report/models/product_product.py
StefanRijnhart: So `self.env.company.currency_id.compare_amounts`?

stock_account_valuation_report/models/product_product.py
StefanRijnhart: `pp.uom_id.rounding`?

stock_account_valuation_report/models/product_product.py
StefanRijnhart: See new commit

stock_account_valuation_report/models/product_product.py
StefanRijnhart: See new commit

gdgellatly (approved):
LGTM, static review

stock_account_valuation_report/models/product_product.py
gdgellatly: @StefanRijnhart The reason I would not use UoM id rounding and use UoM decimal precision is 1 performance but 2, UoM granularity can change, different UoM's can be used etc, and the differences can be below that. However no matter what, the stored UoM precision is going to be identical in the quantity fields. But this point is minor.

AaronHForgeFlow (approved):
Functional review and ok with the changes in code

rousseldenis (approved):
LGTM

PR #371, opened 2025-02-19
[14.0] [ADD] stock_negative_alert_by_email: manage negative in stock via email alert

Borruso:


Borruso:
what do you think? @pedrobaeza

pedrobaeza:
The name should start with `stock`. For example, `stock_negative_alert_by_email` or similar.

Borruso:
> The name should start with `stock`. For example, `stock_negative_alert_by_email` or similar.

okay, thanks now I change

PR #374, opened 2025-02-27
[17.0][ADD] stock_move_pivot_total_price: New module stock_move_pivot_total_price

BernatObrador:
This module introduces **Product total price** for every stock move, wich can be selected as a mesure in Move Analysis.

1. Go to Inventory -> Reporting -> Move Analysis
2. Now in Measures you will have a new field **Total Price**
3. Once this measure is selected the total price of the products based on the quantities and Sales price will be displayed in the pivot view.

cc https://github.com/APSL 167619

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @ppyczko please review


peluko00 (approved):
LGTM, code review and tested in runboat

stock_move_pivot_total_price.zip
ppyczko, line 1: Remove this file please

BernatObrador: Done!! Thanks for your comment.

ppyczko (approved):
LGTM, code review and tested in runboat

mpascuall (approved):
LGTM, tested in runboat

stock_move_pivot_total_price/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
...
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    product_total_price = fields.Float(
+        compute="_compute_product_total_price",
+        string="Total Product Price",
+        store=True,
+    )
+
+    @api.depends("product_id", "quantity")
```
gdgellatly, line 15: list_price is the wrong field, you should be using lst_price. lst_price is a non stored computed field that takes into account list_price and any price_extra. 

EDIT: I did have a note that you would also need to depend on those, but that is crazy, otherwise changing a lst price would update all the past records, so the current dependencies are correct.

stock_move_pivot_total_price/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
...
+    def _compute_product_total_price(self):
+        for record in self:
+            if record.product_id and record.quantity:
```
gdgellatly: This if else is not really necessary. Both fields are required fields. It is OK, just extra processing on a busy table, but maybe it works out faster anyway if lots of 0's.

stock_move_pivot_total_price/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2024 Bernat Obrador
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    product_total_price = fields.Float(
```
gdgellatly, line 9: This module is going to need a pre-init hook, it will be almost uninstallable on large databases. The pre-init hook will need to firstly create the column, and then populate it. Population will be less than trivial having to take account of list_price, and then the sum of extra prices from attribute values.

gdgellatly (changes requested):
Thanks for you contribution, I have added some comments where I feel we have issue.

BernatObrador:
Hi @gdgellatly,

Thank you for your time and feedback.

Your suggestions make perfect sense, so I’ve gone ahead and implemented them. Could you kindly review it again when you have a moment?

stock_move_pivot_total_price/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module introduces Product total price for every stock move, wich can be selected as a mesure in Move Analysis.
```
gdgellatly:
```suggestion
This module introduces Product total price for every stock move, which can be selected as a measure in Move Analysis.
```

gdgellatly (approved):
Aside typos LGTM, static review

lbarry-apsl (approved):
LGTM, code and runboat reviewed

BernatObrador:
Hi @rousseldenis, sorry to ping you directly, just wanted to check if this is ready to be merged. Thanks in advance!

rousseldenis:
/ocabot merge nobump

@BernatObrador I want to warn you about possible performances problems you can have after install of this.

BernatObrador:
Thanks for the comment @rousseldenis.
Could you tell me a bit more about the potential performance issues?
I’ve got this running on a DB with ~93k stock moves and haven’t seen problems, but I’m curious to understand it better.

rousseldenis:
> Thanks for the comment @rousseldenis. Could you tell me a bit more about the potential performance issues? I’ve got this running on a DB with ~93k stock moves and haven’t seen problems, but I’m curious to understand it better.

It's not at the read step but at move creation. As list price on product can be tough to compute if you have complex list prices.

PR #375, opened 2025-03-04
[FIX] [16.0] stock_picking_report_product_sticker: allow stock users to print stickers

Shide:
Since the domain inside `_compute_sticker_ids` requires access to the `ir.model` model, only admins could print stickers. Other users were getting an `AccessError`.
Now tests are executed with an stock user. Running them like this without the fix triggers the error.

Related PR: https://github.com/OCA/account-invoice-reporting/pull/360

MT-9304 @moduon @yajo @rafaelbn @fcvalgar please review if you want 😄


fcvalgar (approved):
LGTM, thank you @Shide

rafaelbn (approved):
👍🏼 

/ocabot merge patch

PR #377, opened 2025-03-05
[16.0][IMP] stock_account_valuation_discrepancy_adjust: Improve code structure

gdgellatly:
Firstly I love this module, I only just discovered it. I just found that the action_create_adjustment didn't separate the imperative parts from the business logic part very well restriciting use. As an example, there is no possibility to extend the creation dictionaries, change move names or determine conditions for adjustment (which should be a property of the product).

As an example, it may be the case where discrepancy is 0, but you are valuing stock by warehouse and it is +1, -1. In that case you will need 2 product lines and 2 counterpart lines.

A simpler example is you just need to add say an operating unit on the move or move lines.

- Factor out move data preparation into separate methods
- Split debit/credit line preparation into dedicated methods
- Fix mismatch between company_id on default journal and company_id on record causing no default journal in multicompany sometimes.

EDIT: Also fix wrong comparison bug in values, this should be done using the currency rounding and not the uom. Otherwise discrepancies remain as small values if uom rounding is less than currency.
EDIT2: Further usability testing reveals a need to be able set user-defined defaults by company. Added.
EDIT3: Fix weird decimal precision choices of Product Price for quantity and discount for value.

### DIFF
--- a/stock_account_valuation_discrepancy_adjust/__init__.py
+++ b/stock_account_valuation_discrepancy_adjust/__init__.py
@@ -1,4 +1,5 @@
 # Copyright 2021 ForgeFlow S.L.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+from . import models
 from . import wizards
--- a/stock_account_valuation_discrepancy_adjust/models/__init__.py
+++ b/stock_account_valuation_discrepancy_adjust/models/__init__.py
@@ -1,0 +1,1 @@
+from . import product_product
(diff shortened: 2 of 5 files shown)


AaronHForgeFlow (approved):
Thank you for this improvement.

I tested with different dates and it worked, and code LGTM so I think we can move forward with this.

gdgellatly:
Sorry I made a mistake here https://github.com/OCA/stock-logistics-reporting/pull/377/files#diff-d6055728635c3de730f384efe25ffbea9ac93cf0417f7337efb3b2a7fb88510e should be seperate PR, will sort tomorrow

gdgellatly:
This is now resolved, extra files removed and on PR #379

giarve (approved):

gdgellatly:
Closing, people on old versions don't want changes and over a year now, maybe better suited to later version

PR #379, opened 2025-03-06
[16.0][FIX] stock_account_valuation_report: decimal precision used in compa…

gdgellatly:
…risons

This PR changes the decimal precisions used in comparisons to values associated with the things being compared.

### DIFF
--- a/stock_account_valuation_report/models/product_product.py
+++ b/stock_account_valuation_report/models/product_product.py
@@ -54,7 +54,7 @@ class ProductProduct(models.Model):
                 ("type", "=", "product"),
             ]
         )
-        dp = self.env["decimal.precision"].precision_get("Product Price")
+        dp = self.env.ref("product.decimal_product_uom").digits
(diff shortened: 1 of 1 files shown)


gdgellatly:
FFS, I had some other test code in here for performance, will fix up

AaronHForgeFlow (approved):
:+1:  LGTM!

giarve (approved):
LGTM

There is still another issue with currencies.

For "Accounting Valuation" it should take the Cost currency also otherwise it will not work properly in multi-company environment with different company currencies

https://github.com/OCA/stock-logistics-reporting/blob/56f7fc922cc9b4efd1100f6a51744871c1201704/stock_account_valuation_report/views/product_product_views.xml#L18

should be added also next to

https://github.com/OCA/stock-logistics-reporting/blob/56f7fc922cc9b4efd1100f6a51744871c1201704/stock_account_valuation_report/views/product_product_views.xml#L36

I might create a PR for it

PR #396, opened 2025-03-10
[18.0][MIG] stock_picking_report_internal_delivery_address: Migration to 18.0

BhaveshHeliconia:


rousseldenis:
/ocabot migration stock_picking_report_internal_delivery_address

stock_picking_report_internal_delivery_address/models/stock_picking.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _is_to_external_location(self):
+        self.ensure_one()
+        return (
```
rousseldenis: A little test for that should be great.

rousseldenis (approved):

BhaveshHeliconia:
@rousseldenis, Please Review.Thanks for the suggestions!

jbaudoux (approved):

david-banon-tecnativa (approved):
TT54521

stock_picking_report_internal_delivery_address/models/stock_picking.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _is_to_external_location(self):
```
pedrobaeza: This method may be used or other usages, so changing it only for getting an address shown seems a bad technique for me. You should instead override `should_print_delivery_address`:

https://github.com/OCA/OCB/blob/76076388cbb4d7e9a5f95b760ae4a0f3786d2fb3/addons/stock/models/stock_picking.py#L1161

stock_picking_report_internal_delivery_address/tests/test_stock_picking_report_internal_delivery_address.py
```diff
@@ -0,0 +1,46 @@
+from odoo.tests.common import TransactionCase
+
+
+class TestStockPicking(TransactionCase):
```
pedrobaeza: This test, for being effective of the final goal of this module, should print the delivery slip, and check that the delivery address is included in it. You can get it as HTML like in https://github.com/OCA/account-invoice-reporting/blob/4e5c239911f49535b0a5506c2d504a07a76227d1/account_invoice_report_grouped_by_picking/tests/test_account_invoice_group_picking.py#L127, and then directly look for the text in the string (instead of parsing it with `lxml` and look for the node, that seems overkill for this case if you put a very especial delivery address to make sure there's no other match in the whole HTML text).

pedrobaeza (changes requested):

BhaveshHeliconia:
@pedrobaeza : Please review. Thanks

stock_picking_report_internal_delivery_address/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def should_print_delivery_address(self):
```
pedrobaeza, line 10: You should still call `super` on all the cases except the specific one we want to catch.

stock_picking_report_internal_delivery_address/tests/test_stock_picking_report_internal_delivery_address.py
```diff
@@ -0,0 +1,129 @@
...
+            {
+                "name": "Test Internal Picking",
+                "picking_type_id": cls.env.ref("stock.picking_type_internal").id,
+                "location_id": cls.env.ref("stock.stock_location_stock").id,
+                "location_dest_id": cls.env.ref("stock.stock_location_output").id,
+                "partner_id": cls.partner.id,
+            }
+        )
+        cls.internal_picking.picking_type_code = "internal"
+
```
pedrobaeza: Don't add empty lines inside a method.

stock_picking_report_internal_delivery_address/tests/test_stock_picking_report_internal_delivery_address.py
```diff
@@ -0,0 +1,129 @@
...
+        cls.env["stock.move"].create(
```
pedrobaeza: Why creating moves apart from the picking? Do it in one shot. This sounds like the direct IA output... Please don't use IA for generating a lot of crap code.

BhaveshHeliconia:
@pedrobaeza : Thank you for the feedback. I've addressed all the comments and made the necessary changes. Please review when you have a moment.

stock_picking_report_internal_delivery_address/tests/test_stock_picking_report_internal_delivery_address.py
```diff
@@ -0,0 +1,132 @@
...
+
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class TestStockPicking(BaseCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+
```
pedrobaeza: Don't add empty lines inside a method.

stock_picking_report_internal_delivery_address/tests/test_stock_picking_report_internal_delivery_address.py
```diff
@@ -0,0 +1,132 @@
...
+                "country_id": cls.env.ref("base.us").id,
+            }
+        )
+
+        cls.product = cls.env["product.product"].create(
+            {
+                "default_code": "AVG",
+                "name": "Average Ice Cream",
+                "is_storable": True,
+                "list_price": 100.0,
```
pedrobaeza: Reduce the number of lines not setting fields that don't add value to the test.

stock_picking_report_internal_delivery_address/tests/test_stock_picking_report_internal_delivery_address.py
```diff
@@ -0,0 +1,132 @@
...
+        cls.outgoing_picking = cls.env["stock.picking"].create(
+            {
+                "name": "Test Outgoing Picking",
+                "picking_type_id": cls.env.ref("stock.picking_type_out").id,
+                "location_id": cls.env.ref("stock.stock_location_stock").id,
+                "location_dest_id": cls.env.ref("stock.stock_location_customers").id,
+                "partner_id": cls.partner.id,
+            }
+        )
+        cls.outgoing_picking.picking_type_code = "outgoing"
```
pedrobaeza: Changing the type of the picking after its creation is not correct.

stock_picking_report_internal_delivery_address/tests/test_stock_picking_report_internal_delivery_address.py
```diff
@@ -0,0 +1,132 @@
...
+
+        cls.env["stock.move"].create(
```
pedrobaeza: You are still creating moves outside the picking.

stock_picking_report_internal_delivery_address/tests/test_stock_picking_report_internal_delivery_address.py
```diff
@@ -0,0 +1,132 @@
...
+                "product_uom": cls.product.uom_id.id,
+                "location_id": cls.outgoing_picking.location_id.id,
+                "location_dest_id": cls.outgoing_picking.location_dest_id.id,
+                "picking_id": cls.outgoing_picking.id,
+            }
+        )
+
+    def test_should_print_delivery_address_internal(self):
+        self.assertTrue(
+            self.internal_picking.should_print_delivery_address(),
```
pedrobaeza: Checking directly a method doesn't have value. You have to test the flow where this call is made.

pedrobaeza (commented):
@BhaveshHeliconia most of my previous comments still apply. Please don't disturb if you haven't solved all of them.

BhaveshHeliconia:
@pedrobaeza : Thank you for taking the time to review, and apologies for wasting your time earlier. I’ve now addressed all your previous comments. Please review when you get a chance, and let me know if anything is still missing.

pedrobaeza (approved):
/ocabot merge nobump

PR #401, opened 2025-03-18
[IMP] stock_average_daily_sale: change order of columns in the report

twalter-c2c:
After using the module for some time, we find the column order not most readable/logical. Here you can find a suggested order.

I added also a possibility to hide/show most of the columns.

PR #355, opened 2024-12-09
[FIX] stock_average_daily_sale: include weekends in demo data

twalter-c2c:
In the test modified by the commit [fc5e8bcdfaf08c06b7f10e6ab7464e330ec86162](https://github.com/OCA/stock-logistics-reporting/commit/fc5e8bcdfaf08c06b7f10e6ab7464e330ec86162), the stock.move is created with yesterday's date. In the demo data, weekends were excluded from the calculation, which resulted in the test failing sometimes (on Sundays and Mondays). Including weekends in the demo config solves this issue.

PR #403, opened 2025-04-04
[16.0][IMP][REF] stock_average_daily_sale: refactor

jbaudoux:
- Allow to set a location from which consumption are computed. This allows to compute consumption on a picking zone instead of the complete warehouse.
- Properly compute the average daily quantity. The old query was lacking a time basis and not computing on daily basis.
- Add max daily consumption metric
- Remove stock level, this is not relevant for this report
- Drop spike exclusion as it was counter-productive the way it was used
- When average is computed on 5 days a week, don't loose any movement happening during weekend
- Added multi-company rules

performance: more than 1000 lines computed per second


jbaudoux:
Tests are red due to network issue

jbaudoux:
For review @rousseldenis @twalter-c2c @lmignon

stock_average_daily_sale/demo/stock_average_daily_sale_config.xml
```diff
@@ -9,7 +9,6 @@
         <field name="abc_classification_level">a</field>
         <field name="period_value">2</field>
         <field name="period_name">week</field>
-        <field name="standard_deviation_exclude_factor">3</field>
         <field name="safety_factor">0.3</field>
         <field name="number_days_qty_in_stock">2</field>
         <field name="exclude_weekends">1</field>
```
twalter-c2c, line 14:
```suggestion
        <field name="exclude_weekends">0</field>
```
Maybe we can include weekends in demo data to be consistent with the change merged in v14 #355.

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -24,26 +27,29 @@ class StockAverageDailySale(models.Model):
...
         digits="Product Unit of Measure",
         required=True,
         help="The quantity delivered on average on one day for this product on "
-        "the period. The spikes are excluded from the average computation.",
+        "the period.",
+    )
+    max_daily_qty = fields.Float(
+        digits="Product Unit of Measure",
+        required=True,
+        help="The max quantity delivered on one day for this product on " "the period.",
```
twalter-c2c:
```suggestion
        help="The max quantity delivered on one day for this product on the period.",
```

stock_average_daily_sale/demo/stock_average_daily_sale_config.xml
jbaudoux: I don't rely much on demo data anymore in the tests. So doesn't really matter

stock_average_daily_sale/views/stock_average_daily_sale.xml
```diff
@@ -42,27 +50,52 @@
...
+                    </group>
+                    <group>
+                        <field name="nbr_sales" />
+                        <field name="average_qty_by_sale" />
+                        <field name="average_daily_sales_count" />
+                        <field name="max_daily_qty" />
+                        <field name="average_daily_qty" />
+                        <field name="daily_standard_deviation" />
+                        <field name="safety" />
+                        <field name="recommended_qty" />
```
twalter-c2c: I suggested a different order of columns in #401 . Hard to say which one is better. To be discussed.

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -164,165 +182,135 @@ def _create_materialized_view(self):
...
-                        JOIN product_product pp on pp.id = sm.product_id
-                        JOIN product_template pt on pp.product_tmpl_id = pt.id
-                        JOIN cfg on cfg.abc_classification_level = coalesce(pt.abc_storage, 'c')
+                        JOIN product_product pp ON pp.id = sm.product_id
+                        JOIN product_template pt ON pp.product_tmpl_id = pt.id
+                        CROSS JOIN LATERAL (
+                            SElECT * FROM cfg
+                            WHERE cfg.abc_classification_level = COALESCE(pt.abc_storage, 'c')
+                            AND sl_src.parent_path ilike concat(cfg.location_parent_path, '%%')
+                            AND sl_dest.parent_path not ilike concat(cfg.location_parent_path, '%%')
```
twalter-c2c, line 232: I wonder if we want to take internal transfers into account. Which we do now, am I right?

twalter-c2c (approved):
Great job! I dropped few minor comments.

stock_average_daily_sale/models/stock_average_daily_sale.py
jbaudoux: Yes it's on purpose to be able to monitor a picking zone instead of stock to customer

stock_average_daily_sale/views/stock_average_daily_sale.xml
jbaudoux: Ah thanks for pointing out. I'll add the optional  here. Seems a nice addition.
For the order of columns, it's a matter of taste... I tried to be logical. Anyway it's a detail

jbaudoux: I did this
https://github.com/OCA/stock-logistics-reporting/pull/403/commits/d5add518e35110eb0acb1a854a24a0fc7766bfb4
based on your suggestion

lmignon (approved):
The readability is highly improved in addition to all the improvements.  Thank you @jbaudoux . LGTM (Code review).

As a side note,  I'm a little alarmed by the deletion of the exclusion of peaks, which was intended to ensure that the average does not take into account “exceptional events”, which should not be taken into account when talking about averages. (we'll see..)

PR #407, opened 2025-04-17
[17.0][MIG] stock_quant_history: Migration to 17.0

nico-foodles:


stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -69,7 +70,16 @@ def _compute_name(self):
         dt_format = lang.date_format + " " + lang.time_format
 
         for rec in self:
-            rec.name = _("Snapshot %s") % (rec.inventory_date.strftime(dt_format))
+            if not rec.inventory_date:
```
petrus-v, line 73: *question*: this field is required do we needs to prevent this ? is it because there are no default and code is triggered on creation ?

stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -69,7 +70,16 @@ def _compute_name(self):
...
+                rec.name = _("Snapshot")
+                continue
+
+            user_tz = self.env.user.tz or "UTC"
+            user_timezone = timezone(user_tz)
```
petrus-v, line 78: Is there a timezone on warehouse, I've the feeling the warehouse timezone would be more accurate than the  user timezone ?

if not I'm fine with that code

petrus-v (approved):

stock_quant_history/models/stock_quant_history_snapshot.py
nico-foodles: This piece of code exists because when you try to create a snapshot via the interface, the name is generated with a non-existent date and raises an error. I've also tried using a default date but found it more elegant not to use one as the name is then overwritten.

stock_quant_history/models/stock_quant_history_snapshot.py
nico-foodles: After checking I couldn't find any, maybe I'm wrong

barkat-matthias (approved):

stock_quant_history/models/stock_quant_history_snapshot.py
petrus-v: ok, thanks for double check, anyway better this way

etobella (approved):
/ocabot merge nobump

PR #408, opened 2025-04-17
[17.0][MIG] stock_quant_history_queued: Migration to 17.0

nico-foodles:


barkat-matthias (approved):

.direnv/CACHEDIR.TAG
```diff
@@ -0,0 +1,4 @@
+Signature: 8a477f597d28d172789f06886806bc55
```
petrus-v: could you rewrite the history that those files are never present in the git history ?

petrus-v (commented):
Migration looks good to me but could you remove unwanted files from the history ?

.direnv/CACHEDIR.TAG
nico-foodles: Done

petrus-v (approved):

rousseldenis (approved):

PR #414, opened 2025-05-21
[14.0][IMP] stock_quant_history

StephaneMangin:
Add a constraint on locked picking while an history snapshot is generated.

Included a light refactoring of tests suites.

Propagated to [17.0 PR](https://github.com/OCA/stock-logistics-reporting/pull/415)

### DIFF
--- a/stock_quant_history/__manifest__.py
+++ b/stock_quant_history/__manifest__.py
@@ -4,17 +4,20 @@
 {
     "name": "Stock Quant History",
     "summary": "Re-generate stock quants for given date",
-    "version": "14.0.1.0.0",
+    "version": "14.0.1.1.0",
     "license": "AGPL-3",
-    "author": "Pierre Verkest <pierreverkest84@gmail.com>, Odoo Community Association (OCA)",
+    "author": "Pierre Verkest <pierreverkest84@gmail.com>, Stéphane Mangin "
+    "<stephane.mangin@foodles.com>, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-reporting",
     "depends": ["stock"],
     "maintainers": [
         "petrus-v",
+        "StephaneMangin",
     ],
     "data": [
         "security/ir.model.access.csv",
         "views/stock-quant-history-snapshot.xml",
         "views/stock-quant-history.xml",
+        "views/res_config_settings_views.xml",
     ],
 }
--- a/stock_quant_history/models/stock_picking.py
+++ b/stock_quant_history/models/stock_picking.py
@@ -1,0 +1,59 @@
+# Copyright 2025 Foodles (https://www.foodles.co/).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def get_stock_quant_history(self):
+        """Get the last stock quant history related to this picking.
+        The picking must be done.
+        :return: stock.quant.history recordset"""
+        self.ensure_one()
+        history_model = self.env["stock.quant.history"]
+        if self.state == "done" and self.date_done:
+            domain = [
+                ("inventory_date", ">=", self.date_done),
+                ("product_id", "in", self.move_line_ids.mapped("product_id").ids),
+                ("lot_id", "in", self.move_line_ids.mapped("lot_id").ids),
+            ]
+            return history_model.search(domain)
+        return history_model
+
+    def action_toggle_is_locked(self):
+        # Ensure the picking is allowed to be unlocked before toggling the lock status.
+        self.check_unlock_allowed()
+        return super().action_toggle_is_locked()
+
+    def check_unlock_allowed(self):
+        """Check if the picking can be unlocked.
+        :return: True if the picking can be unlocked
+        :raises ValidationError: if not allowed to unlock"""
+        separator = "\n - "
+        self.ensure_one()
+        if self.is_locked:
+            history_lines = self.get_stock_quant_history()
+            if history_lines.exists():
+                snapshot_text = separator.join(
+                    history_lines.mapped("snapshot_id").mapped("display_name")
+                )
+                raise ValidationError(
+                    _("You cannot unlock '%s' as stock histories exists:%s%s.")
+                    % (self.display_name, separator, snapshot_text)
+                )
+        return True
+
+    def write(self, vals):
+        # Constraint the lock status to avoid unlocking a record when a related quant
+        # history exists.
+        # An unlock action is propagated
+        if "is_locked" in vals and not vals.get("is_locked"):
+            # Reducing to locked status and done picking
+            for picking in self.filtered(
+                lambda pick: pick.is_locked and pick.state == "done" and pick.date_done
+            ):
+                picking.check_unlock_allowed()
+        return super().write(vals)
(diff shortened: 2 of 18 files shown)


stock_quant_history/__manifest__.py
```diff
@@ -4,13 +4,15 @@
...
+    "version": "14.0.1.1.0",
     "license": "AGPL-3",
-    "author": "Pierre Verkest <pierreverkest84@gmail.com>, Odoo Community Association (OCA)",
+    "author": "Pierre Verkest <pierreverkest84@gmail.com>, Foodles, Stéphane Mangin "
+    "<stephane.mangin@foodles.com>, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-reporting",
     "depends": ["stock"],
     "maintainers": [
         "petrus-v",
+        "Foodles",
```
petrus-v: Not sure Foodles is a valid github user account, this gives merge access to the user on any PR related to that module.

If you want to add yourself, you are welcome to become maintainer of this module with me, as you wish !

petrus-v (approved):

petrus-v (approved):

stock_quant_history/tests/test_stock_picking.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2025 Foodles (https://www.foodles.co/).
```
petrus-v, line 1: In order test can be launch by odoo launcher you should import this file (which should fix the coverage I belives)

stock_quant_history/readme/DESCRIPTION.rst
```diff
@@ -7,3 +7,5 @@ are present in the database.
 
 Next snapshot is computed based on the previous snapshot present in the database.
 
+Once a snapshot has been taken, every locked and done picking won't be able to unlock
+to avoid incoherent stock history.
```
petrus-v, line 11: While creating new snapshot should re-lock un-lock picking ?

I believe it should be at least in the in the Roadmap, there is nothing to force user to re-lock picking once it has been unlocked.

alexandregaldeano: I agree that unlocked picking should be locked on snapshot creation :)

stock_quant_history/models/stock_picking.py
```diff
@@ -0,0 +1,57 @@
...
+                    _("You cannot unlock '%s' as stock histories exists:%s%s.")
+                    % (self.display_name, separator, snapshot_text)
+                )
+        return True
+
+    def write(self, vals):
+        # Constraint the lock status to avoid unlocking a record when a related quant
+        # history exists.
+        # An unlock action is propagated
+        if "is_locked" in vals and not vals.get("is_locked"):
```
alexandregaldeano, line 53: **nitpick, suggestion**: invert `if` + early `return`?

alexandregaldeano (approved):
Once @petrus-v 's concerns are addressed, LGTM

stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -182,7 +184,9 @@ def _generate_stock_quant_history(self):
                     ),
                     precision_rounding=move_line.product_id.uom_id.rounding,
                 )
-
+                pickings |= move_line.picking_id
+        # Lock all related pickings
+        pickings.write({"is_locked": True})
```
petrus-v: I'm wondering if odoo perform write on record where is_locked is already True other wise I would probably filter before write:

```suggestion
        pickings.filtered(lambda picking: not picking.is_locked).write({"is_locked": True})
```

StephaneMangin: Value check is performed here in fields.py
```
def get_records_different_from(self, records, field, value):
        """ Return the subset of ``records`` that has not ``value`` for ``field``. """
        field_cache = self._get_field_cache(records, field)
        if field.translate:
            lang = records.env.lang or 'en_US'

            def get_value(id_):
                cache_value = field_cache[id_]
                return None if cache_value is None else cache_value[lang]
        else:
            get_value = field_cache.__getitem__

        ids = []
        for record_id in records._ids:
            try:
                val = get_value(record_id)
            except KeyError:
                ids.append(record_id)
            else:
                if field.type == "monetary":
                    value = field.convert_to_cache(value, records.browse(record_id))
                if val != value:
                    ids.append(record_id)
        return records.browse(ids)
```

So answer is yes, but record by records even in a write with multiple records.

stock_quant_history/models/stock_picking.py
StephaneMangin: I choose to be explicit for clarity. I don't see any gain to early return the super if not in this usecase. Do you see anything else ?

stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -151,38 +151,28 @@ def _generate_stock_quant_history(self):
         _logger.info(
             "Apply %s stock.move.line since previous snapshot", len(stock_move_lines)
         )
+        pickings = self.env["stock.picking"]
         ignored_location_usage = self._ignored_location_usage()
         for move_line in stock_move_lines:
+            quant_copy = quant_history[
+                (move_line.product_id, move_line.lot_id, move_line.location_id)
```
petrus-v: surprised that test do not failed probably we can improve but depending of the condition we use the location_id or location_dest_id. I think only quantity can be refactored

stock_quant_history/models/stock_picking.py
```diff
@@ -0,0 +1,56 @@
...
+        """Check if the picking can be unlocked.
+        :return: True if the picking can be unlocked
+        :raises ValidationError: if not allowed to unlock"""
+        separator = "\n - "
+        self.ensure_one()
+        if self.is_locked:
+            history_lines = self.get_stock_quant_history()
+            if history_lines.exists():
+                snapshot_text = separator.join(
+                    history_lines.mapped("snapshot_id").mapped("display_name")
```
petrus-v, line 41: non blocking:

```suggestion
                    history_lines.snapshot_id.mapped("display_name")
```

stock_quant_history/models/stock_picking.py
```diff
@@ -0,0 +1,56 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def get_stock_quant_history(self):
+        """Get the last stock quant history related to this picking.
+        :return: stock.quant.history recordset"""
+        self.ensure_one()
+        history_model = self.env["stock.quant.history"]
+        domain = [
+            ("inventory_date", ">=", max(self.date_done, self.date)),
```
petrus-v: Do we prevent date_done to be different to False? I'm wondering if this code would raise error

```
>>> from datetime import datetime
>>> max(datetime.now(), False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'bool' and 'datetime.datetime'
```

stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -151,38 +151,31 @@ def _generate_stock_quant_history(self):
         _logger.info(
             "Apply %s stock.move.line since previous snapshot", len(stock_move_lines)
         )
+        pickings = self.env["stock.picking"]
         ignored_location_usage = self._ignored_location_usage()
         for move_line in stock_move_lines:
+            quantity = tools.float_round(
```
petrus-v: not sure the impacted rounding subtraction result vs rounding and subtract

PR #420, opened 2025-07-03
[14.0][FIX] stock_picking_report_valued: prevent singleton error in `res.currency`

carlos-lopez-tecnativa:
for warehouses with multiple delivery steps

This report adds new fields of `Monetary` type. When the report is printed, Odoo tries to access `currency_id`, which is related to the `sale_line`. This field is only set on the final delivery step.

However, when the warehouse is configured with two or three delivery steps, the initial pickings contain move lines that are not linked to any sale line, which raises the following error: `ValueError: Expected singleton: res.currency()`

This commit ensures that the fields are rendered only if at least one move line is linked to a sale line, thus preventing the error.

TT50920
@Tecnativa @pedrobaeza @carlosdauden @CarlosRoca13 could you please review this?

### DIFF
--- a/stock_picking_report_valued/report/stock_picking_report_valued.xml
+++ b/stock_picking_report_valued/report/stock_picking_report_valued.xml
@@ -94,12 +94,15 @@
                 <td class="text-right" groups="product.group_discount_per_so_line">
                     <span t-field="move_line.sale_discount" />
                 </td>
-                <td class="text-right"><span
+                <td class="text-right">
+                    <span
                         t-field="move_line.sale_price_subtotal"
-                    /></td>
-                <td class="text-right"><span
-                        t-field="move_line.sale_tax_description"
-                    /></td>
+                        t-options='{"display_currency": o.currency_id}'
+                    />
+                </td>
+                <td class="text-right">
+                    <span t-field="move_line.sale_tax_description" />
+                </td>
             </t>
         </xpath>
     </template>


CarlosRoca13:
Using a method like [this](https://github.com/odoo/odoo/blob/e9bd1cb33bc26cb3153c72dd2319f5f9a9bb82da/addons/stock/report/report_stock_forecasted.py#L135-L140) to take the final move is not possible?

In v15 this method is defined on the `stock_move`, but we can do something like `stock` module does with `report_stock_forecasted`. What do you think?

carlos-lopez-tecnativa:
> Using a method like [this](https://github.com/odoo/odoo/blob/e9bd1cb33bc26cb3153c72dd2319f5f9a9bb82da/addons/stock/report/report_stock_forecasted.py#L135-L140) to take the final move is not possible?
> 
> In v15 this method is defined on the `stock_move`, but we can do something like `stock` module does with `report_stock_forecasted`. What do you think?

I think your suggestion is complicated to implement. Currently, the [sale_line](https://github.com/OCA/stock-logistics-reporting/blob/ec9a3a94fd43e9d1961fe05a4f4b9f70aa1fef07/stock_picking_report_valued/models/stock_move_line.py#L14C5-L14C14) field is a related field from `move_id`, but Odoo only fills `sale_line_id` on the `stock.move` of the final step. So, this field is empty in the earlier pickings.

Making the change only at the report level would result in the fields being displayed but showing zero values, because [this line](https://github.com/OCA/stock-logistics-reporting/blob/ec9a3a94fd43e9d1961fe05a4f4b9f70aa1fef07/stock_picking_report_valued/models/stock_move_line.py#L63-L65) skips computation if the field is empty.

To implement it correctly, I would need to change the `sale_line` field from `related` to a `computed` field and retrieve the sale line across all related `stock.move` records.

@pedrobaeza @carlosdauden What do you think? Is it worth making this change, or is it enough to just hide the field on intermediate pickings (since it works correctly in the final picking)?

pedrobaeza:
I think the fields should be rendered as well, but with 0. You can default the currency to any value instead.

carlos-lopez-tecnativa:
I updated the code to keep the changes minimal and prevent the error. The fields are now rendered in the report but display zero values.

@pedrobaeza @CarlosRoca13, please update your review and let me know if this is OK for you.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #423, opened 2025-07-11
[16.0][IMP] stock_quantity_history_location: invisible location except from stock.quant view

AungKoKoLin1997:
This PR makes the location field invisible when the model is not stock.quant (locations view), because the location context is not used in other models (e.g., `stock.valuation.layer`) and choosing location will not effect the functionality. So, displaying this field in the pop-up of other models (e.g., Stock Valuation Layer) may confuse users.

@qrtl QT4756

### DIFF
--- a/stock_quantity_history_location/tests/test_stock_quantity_history_location.py
+++ b/stock_quantity_history_location/tests/test_stock_quantity_history_location.py
@@ -31,35 +31,41 @@ class TestStockQuantityHistoryLocation(TestCommon):
         )
         cls._create_stock_move(cls, location_dest_id=cls.child_test_stock_loc, qty=100)
 
-    def test_01_wizard_past_date(self):
+    def get_stock_quantity_history_action(self, location, to_date):
         wizard = self.env["stock.quantity.history"].create(
             {
-                "location_id": self.test_stock_loc.id,
-                "include_child_locations": True,
-                "inventory_datetime": fields.Datetime.now(),
+                "location_id": location.id,
+                "inventory_datetime": to_date,
             }
         )
-        action = wizard.with_context(company_owned=True).open_at_date()
+        return wizard.open_at_date()
+
+    def test_stock_quantity_history_location(self):
+        current_date = fields.Datetime.now()
+        past_date = fields.Datetime.to_datetime("2019-08-10 00:00:00")
+        action = self.get_stock_quantity_history_action(
+            self.child_test_stock_loc, current_date
+        )
+        self.assertEqual(action["context"].get("to_date"), current_date)
         self.assertEqual(
             self.product.with_context(**action["context"]).qty_available, 100.0
         )
+        action = self.get_stock_quantity_history_action(
+            self.child_test_stock_loc, past_date
+        )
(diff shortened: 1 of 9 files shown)


yostashiro (changes requested):
I think `include_child_locations` should be removed. See https://github.com/odoo/odoo/pull/78078.

Can you please also update the README accordingly, as the usage seems outdated.

AungKoKoLin1997:
> I think `include_child_locations` should be removed. See [odoo/odoo#78078](https://github.com/odoo/odoo/pull/78078).
> 
> Can you please also update the README accordingly, as the usage seems outdated.

@yostashiro Done!

stock_quantity_history_location/tests/test_stock_quantity_history_location.py
```diff
@@ -31,35 +31,43 @@ def setUpClass(cls):
...
+
+    def test_stock_quantity_history_location(self):
+        current_date = fields.Datetime.now()
+        past_date = fields.Datetime.to_datetime("2019-08-10 00:00:00")
+        action = self.get_stock_quantity_history_action(
+            self.child_test_stock_loc, current_date
+        )
+        to_date_in_context = action["context"].get("to_date")
+        self.assertIsNotNone(to_date_in_context)
+        self.assertEqual(to_date_in_context, current_date)
```
yostashiro: Non-blocking. I think this should be sufficient.
```suggestion
        self.assertEqual(action["context"].get("to_date"), current_date)
```

yostashiro (approved):
LGTM.

nobuQuartile (approved):
Functional review
LGTM

AungKoKoLin1997:
@oca/stock-logistics-reporting-maintainers
Can we get this PR merged?

stock_quantity_history_location/wizards/stock_quantity_history.xml
```diff
@@ -7,11 +7,10 @@
         <field name="inherit_id" ref="stock.view_stock_quantity_history" />
         <field name="arch" type="xml">
             <field name="inventory_datetime" position="before">
-                <field name="location_id" groups="stock.group_stock_multi_locations" />
                 <field
-                    name="include_child_locations"
-                    attrs="{'invisible': [('location_id', '=', False)]}"
+                    name="location_id"
                     groups="stock.group_stock_multi_locations"
+                    invisible="context.get('active_model') != 'stock.quant'"
```
rolandojduartem, line 13: does this work on Odoo 16.0? I think it should be attrs

AungKoKoLin1997: @rolandojduartem 
It works on odoo 16.0. You can see how odoo standard field `inventory_datetime` displays in view depends on model.
https://github.com/odoo/odoo/blob/1adee2b7452c072bcce51cd75641a9e4c2b7f219/addons/stock_account/wizard/stock_quantity_history.xml#L9-L19.

kanda999 (approved):
Functional review: it works in prod server

rousseldenis (approved):

PR #424, opened 2025-07-11
[REF] stock_quant_history: performance improvements

petrus-v:
Creating a new snapshot has 2 main phases that we could optimise in different ways

* duplicate previous snapshot
* compute new stock parsing stock.move.line

## duplicate previous snapshot

This can be at SQL level to avoid ORM overhead creating the request dynamically.

On my laptop duplicate 93633 previous stock.quant.history records goes from 5'21 to 0'09'' (x35 faster)

## Compute new stock

This loop over stock.move.line between previous snapshot and the new one to compute new stock.quant.history. Looping over 2 read_group improve the situation in my current context processing 96462 stock.move.line from 1'37'' to 0'23''.

### DIFF
--- a/stock_quant_history/models/stock_quant_history_snapshot.py
+++ b/stock_quant_history/models/stock_quant_history_snapshot.py
@@ -98,17 +98,7 @@ class StockQuantHistorySnapshot(models.Model):
             "inventory",
         ]
 
-    def _generate_stock_quant_history(self):
-        self.ensure_one()
-        self.generated_date = fields.Datetime.now()
-        previous_quant_snapshot = self.search(
-            [
-                ("state", "=", "generated"),
-                ("inventory_date", "<=", self.inventory_date),
-            ],
-            order="inventory_date desc",
-            limit=1,
-        )
+    def _copy_previous_stock_quant_history(self):
         quant_history = DefaultDict(
             lambda product, lot, location: self.env["stock.quant.history"]
             .sudo()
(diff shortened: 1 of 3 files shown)


alexandregaldeano (commented):
I wasn't expecting this change to be this massive, I'll take a second look tomorrow 😬

stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -122,63 +112,154 @@ def _generate_stock_quant_history(self):
...
                 "Duplicate %s previous stock.quant.history...",
-                len(previous_quant_snapshot.stock_quant_history_ids),
+                len(self.previous_snapshot_id.stock_quant_history_ids),
             )
-            for stock_quant_history in previous_quant_snapshot.stock_quant_history_ids:
-                # copy is around 3x slower than create !
-                quant_copy = quant_history[
+            self.flush()
+            self.env.cr.execute(
+                "INSERT INTO stock_quant_history ("
```
alexandregaldeano, line 129: **suggestion:** Maybe use `f""" """` to improve readability? 🤔

alexandregaldeano (approved):
LGTM

stock_quant_history/models/stock_quant_history_snapshot.py
petrus-v: I did in first try but get upset with linter for un obscure reason...

stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -122,63 +112,154 @@ def _generate_stock_quant_history(self):
                 }
             )
         )
-        self.previous_snapshot_id = previous_quant_snapshot
-
-        _logger.info("Processing %s from %s", self.name, self.previous_snapshot_id.name)
-        if previous_quant_snapshot.stock_quant_history_ids.exists():
+        duplicated_fields = ", ".join(self.env["stock.quant.history"]._fields_to_copy())
```
ltabis, line 115: nitpick: I would suggest renaming this to `fields_to_duplicate`, because on first read I instantly though that this variable stored fields that where duplicated, but it just stores fields that WILL BE duplicated.

ltabis (approved):
👍

stock_quant_history/models/stock_quant_history_snapshot.py
alexandregaldeano: Soooooorrrry, I thought I already did the comment, you can ignore 😬

stock_quant_history/models/stock_quant_history_snapshot.py
petrus-v: oups... I missed your comment directly moving to the bottom of the page.... this is relevant suggestion !

petrus-v: a follow-up is here: https://github.com/OCA/stock-logistics-reporting/pull/430

PR #427, opened 2025-07-11
[16.0][ADD] stock_portal_lot_list_download: New module

pilarvargas-tecnativa:
@Tecnativa TT56946

@pedrobaeza  @victoralmau please review


pedrobaeza (changes requested):
Functional analysis in runboat:

image


And you need to add a column for quantities, as lots may be more than one.

pilarvargas-tecnativa:
> And you need to add a column for quantities, as lots may be more than one.

All done

image

stock_portal_lot_list_download/controllers/portal.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+        # Columns
+        worksheet.set_column("A:A", 40)  # Product
+        worksheet.set_column("B:B", 25)  # Lot
+        worksheet.set_column("C:C", 15)  # Quantity done
+        # Headers
+        worksheet.merge_range("A1:B1", f"Picking: {picking.name}", title_format)
+        worksheet.write("A3", "Product", header_format)
+        worksheet.write("B3", "Lot/Serial Number", header_format)
+        worksheet.write("C3", "Qty Done", header_format)
```
pedrobaeza:
```suggestion
        worksheet.write("C3", "Quantity", header_format)
```

pedrobaeza (commented):
Please include a test. You may just call directly the controller URL from tests a check that something is returned with no error.

stock_portal_lot_list_download/controllers/portal.py
pilarvargas-tecnativa: done, thanks!

pedrobaeza (approved):

stock_portal_lot_list_download/controllers/portal.py
```diff
@@ -0,0 +1,70 @@
...
+        worksheet.write("C3", "Quantity", header_format)
+        # Data
+        row = 3
+        for ml in picking.move_line_ids_without_package:
+            worksheet.write(row, 0, ml.product_id.display_name, cell_format)
+            worksheet.write(row, 1, ml.lot_id.name or "", cell_format)
+            worksheet.write(row, 2, ml.qty_done, cell_format)
+            row += 1
+        workbook.close()
+        output.seek(0)
```
carlos-lopez-tecnativa: Could you please split this function into a new function that only prepares the Excel file and returns it? This would help maintain the _Single Responsibility Principle (SRP)_ and make inheritance easier.

stock_portal_lot_list_download/controllers/portal.py
```diff
@@ -0,0 +1,70 @@
...
+        worksheet.set_column("B:B", 25)  # Lot
+        worksheet.set_column("C:C", 15)  # Quantity done
+        # Headers
+        worksheet.merge_range("A1:B1", f"Picking: {picking.name}", title_format)
+        worksheet.write("A3", "Product", header_format)
+        worksheet.write("B3", "Lot/Serial Number", header_format)
+        worksheet.write("C3", "Quantity", header_format)
+        # Data
+        row = 3
+        for ml in picking.move_line_ids_without_package:
```
carlos-lopez-tecnativa: I think this should only print products with lots or serial numbers, right?
image

carlos-lopez-tecnativa (changes requested):

stock_portal_lot_list_download/controllers/portal.py
pilarvargas-tecnativa: The requirement is to download the delivery slip in Excel, only if the delivery slip contains serial numbers. It does not specify not to print only products with lot or serial numbers.

pedrobaeza: Yeah, but it's reasonable to not include them in this file.

pilarvargas-tecnativa: Done

carlos-lopez-tecnativa (approved):
LGTM

PR #430, opened 2025-07-25
[REF] stock_quant_history: code styling

petrus-v:


stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -227,6 +229,17 @@ def _apply_stock_move_lines_group(
             )
 
     def _apply_stock_move_lines(self, quant_history):
+        _logger.info(
+            "apply %d stock.move.line between from previous "
+            "snapshot (%s) to the current (%s)",
+            self.env["stock.move.line"]
+            .sudo()
+            .search_count(
```
alexandregaldeano, line 237: **question:** Isn't a search count a bit costly just for a logger?

petrus-v: yes it cost but do not represent a very much time according the whole process. I think it's quite a value-able information while doing maintenance task that gives some feedback on what's going on... very debatable, If you prefer I let the logger without the computation?

alexandregaldeano: Good enough for me 😉

alexandregaldeano (approved):

bosd:
Can you please include [14.0] in the title?

PR #440, opened 2025-08-25
[17.0][FIX] stock_picking_report_valued: Fix currency_id field

CarlosRoca13:
We make sure that the currency_id field always has a value to avoid errors. For example, when installing this module with stock_picking_report_delivery_cost and running the tests, there is a case where the picking has no currency defined, which blocks the CI.

FW-PORT of https://github.com/OCA/stock-logistics-reporting/pull/439

cc @Tecnativa 

ping @pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #447, opened 2025-11-06
[18.0][IMP] stock_quantity_history_location: invisible location except from Stock view

AungKoKoLin1997:
Forward port of https://github.com/OCA/stock-logistics-reporting/pull/423

This PR makes the location field invisible when the model is not `product.product` (Stock view), because the location context is not used in other models (e.g., `stock.valuation.layer`) and choosing location will not effect the functionality. So, displaying this field in the pop-up of other models (e.g., Stock Valuation Layer) may confuse users.

@qrtl QT5862


yostashiro (approved):
Code review and functional test. LGTM.

kanda999 (approved):
LGTM

PR #450, opened 2025-11-18
[18.0][FIX] stock_picking_report_valued: wrong prices

david-s73:
- The way the fields were calculated has been modified, since after performing the new() function, it recalculates the fields and loses them, only showing the unit price at the end. Therefore, it has been modified so that if it has a quantity different from the sales order or a different UOM, it first calculates the subtotal price and then performs the new() function. This way, the prices are calculated correctly in the report and the lines.

- I'm leaving screenshots taken on the runboat so you can replicate it.

image

image

image

### DIFF
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -60,7 +60,6 @@ class StockMoveLine(models.Model):
             if not valued_line:
                 continue
             quantity = line._get_report_valued_quantity()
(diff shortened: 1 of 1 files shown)


Reyes4711-S73 (approved):
LGTM

Raul-S73 (approved):

Reyes4711-S73:
@OCA/logistics-maintainers Please, can you review/merge this PR?

pedrobaeza (approved):
/ocabot merge patch

PR #454, opened 2025-11-19
[16.0] [FIX] stock_picking_report_valued_sale_mrp: fix components per kit computation when redelivering

### LINKED ISSUE #451: [16.0] stock_picking_report_valued_sale_mrp: kit component demanded quantity is doubled when a delivery is fully returned and then redelivered
### Module  
`stock_picking_report_valued_sale_mrp` – Odoo 16.0

### Description  
When a kit (phantom BOM) is delivered, fully returned, and then redelivered, the valued delivery reports **Delivery Slip** show **twice the correct quantity and value for kit components**.

The root cause is the method `_get_components_per_kit()` in `models/stock_move.py`, which incorrectly computes the demanded component quantity from the sale line when returns and redeliveries are present.

### Current (wrong) behaviour  
The method returns **80 components per kit** instead of the correct **40 components per kit**

### Expected behaviour  
A sale line demanding **2 kits** must always show **40 components per kit** in the report (80 units delivered / 2 kits ordered = 40), regardless of how many times the delivery was returned and redelivered).  
Net effect: customer finally received 2 kits → component consumption must be valued for 2 kits only.

### Real data example (anonymized but 100% representative)

| id    | name     | product_id | product_qty | product_uom_qty | move_dest_ids   | move_orig_ids   | picking_id | state | origin_returned_move_id                  | returned_move_ids | quantity_done | picking_code | sale_line_id |
|-------|----------|------------|-------------|---------------------|-----------------|-----------------|-----------------|-------|------------------------------------------|-------------------|---------------|--------------|--------------|
| MV-01 | [BOX-01] | PRD-01     | 80          | 80                  | MV-02           | PK-01, ALB/XXXX | PK-10           | done  | false                                    | MV-02             | 80            | outgoing     | SO-2501524   |
| MV-02 | [BOX-01] | PRD-01     | 80          | 80                  | MV-01           | PK-02, RET/XXXX | PK-11           | done  | MV-01 (SO-2501524: Stock>Customers)      | (none)            | 80            | incoming     | SO-2501524   |
| MV-03 | [BOX-01] | PRD-01     | 80          | 80                  | PK-03, ALB/XXXX | (none)          | PK-12           | done  | false                                    | (none)            | 80            | outgoing     | SO-2501524   |

### Why the current code fails
Current implementation (lines 9–31):

```python
component_demand = sum(
    sale_line.move_ids.filtered(
        lambda x: x.product_id == self.product_id
        and not x.origin_returned_move_id   # excludes only return moves
        and ...
    ).mapped("product_uom_qty")
)

Result with the data above:

MV-01 → included (80)
MV-02 → correctly excluded (return)
MV-03 → included (80)

→ total = 160 → 160 / 2 kits ordered = 80 components per kit (doubled)
The original delivery quantity is never reduced by the quantity that was later returned, producing systematic over-valuation every time a return + replacement occurs.

### Buggy code (current)
https://github.com/OCA/stock-logistics-reporting/blob/56f7fc922cc9b4efd1100f6a51744871c1201704/stock_picking_report_valued_sale_mrp/models/stock_move.py#L9-L31

### PR DESCRIPTION
**Pull request by rrebollo**
Fixes #451

This PR contains **two commits**:

1. **Adds a test** that reproduces the issue.  
2. **Introduces a proposed fix** for the issue.

### 🧩 Context

The problem occurs in the following workflow:

1. Sell a kit.  
2. Confirm the Sales Order (SO).  
3. Validate the related picking.  
4. Return the picking.  
5. Cancel the SO.  
6. Set the SO back to draft.  
7. Confirm the SO again.

At this point, the system generate another picking and you validate this too, and we noticed that the **kit component quantities** shown on the **delivery slip report** were incorrect.

### 🔍 Root Cause

The issue appears to stem from the calculation of the **component quantities per kit** during this sequence of operations.

### 🛠️ Proposed Fix

This PR provides an initial attempt to resolve the incorrect quantity computation. Additional details and discussion can be found in the linked issue.

HT14789
@BinhexTeam

### DIFF
--- a/stock_picking_report_valued_sale_mrp/models/stock_move.py
+++ b/stock_picking_report_valued_sale_mrp/models/stock_move.py
@@ -22,6 +22,7 @@ class StockMove(models.Model):
             sale_line.move_ids.filtered(
                 lambda x: x.product_id == self.product_id
                 and not x.origin_returned_move_id
+                and not x.returned_move_ids
                 and (
                     x.state != "cancel"
                     or (x.state == "cancel" and x.picking_id.backorder_id)
--- a/stock_picking_report_valued_sale_mrp/tests/test_stock_picking_report_valued_mrp.py
+++ b/stock_picking_report_valued_sale_mrp/tests/test_stock_picking_report_valued_mrp.py
@@ -45,22 +45,35 @@ class TestStockPickingValuedMrp(TestStockPickingValued):
         cls.product_2 = cls.product_product.create(
             {"name": "Product test 2", "type": "product"}
         )
-        order_form = Form(cls.env["sale.order"])
-        order_form.partner_id = cls.partner
-        with order_form.order_line.new() as line_form:
-            line_form.product_id = cls.product_kit
-            line_form.product_uom_qty = 5
-            line_form.price_unit = 29.9
-            line_form.tax_id.clear()
-            line_form.tax_id.add(cls.tax10)
-        cls.sale_order_3 = order_form.save()
-        cls.sale_order_3.action_confirm()
-        # Maybe other modules create additional lines in the create
-        # method in sale.order model, so let's find the correct line.
-        cls.order_line = cls.sale_order_3.order_line.filtered(
-            lambda r: r.product_id == cls.product_kit
+        (
+            cls.sale_order_3,
+            cls.order_line,
+            cls.order_out_picking,
+        ) = cls._create_sale_order_with_lines(
+            cls,
+            product=cls.product_kit,
+            quantity=5,
+            price_unit=29.9,
+            tax=cls.tax10,
         )
-        cls.order_out_picking = cls.sale_order_3.picking_ids
+
+    def _create_sale_order_with_lines(self, product, quantity, price_unit, tax=None):
+        """Create a sale order with a single line, confirm it, and return
+        the sale order, filtered order line, and picking."""
+        order_form = Form(self.env["sale.order"])
+        order_form.partner_id = self.partner
+        with order_form.order_line.new() as line_form:
+            line_form.product_id = product
+            line_form.product_uom_qty = quantity
+            line_form.price_unit = price_unit
+            line_form.tax_id.clear()
+            if tax:
+                line_form.tax_id.add(tax)
+        sale_order = order_form.save()
+        sale_order.action_confirm()
+        order_line = sale_order.order_line.filtered(lambda r: r.product_id == product)
+        picking = sale_order.picking_ids[0] if sale_order.picking_ids else None
+        return sale_order, order_line, picking
 
     def test_01_picking_confirmed(self):
         for line in self.order_out_picking.move_ids:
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by JordiMForgeFlow — APPROVED** [review]
code LGTM 👍🏼

**Review by GuillemCForgeFlow — APPROVED** [review]
code LGTM 👍🏿

**Comment by rrebollo** [comment]
@rousseldenis could you merge this one?

**Review by BhaveshHeliconia — APPROVED** [review]
LGTM!

**Comment by rrebollo** [comment]
@pedrobaeza could U merge this one?

File: stock_picking_report_valued_sale_mrp/models/stock_move.py [hunk 1]
```diff
@@ -18,14 +18,18 @@ def _get_components_per_kit(self):
...
+        kit_moves = sale_line.move_ids.filtered(
+            lambda x: x.product_id == self.product_id
+            and (
+                x.state != "cancel"
+                or (x.state == "cancel" and x.picking_id.backorder_id)
+            )
+        )
+
+        kit_moves_to_sum = kit_moves.filtered(lambda x: not x.origin_returned_move_id)
+        kit_moves_to_subtract = kit_moves - kit_moves_to_sum
```
pedrobaeza: Instead of complicating this way, just sum the proper moves.

pedrobaeza (changes requested):
Don't introduce extra empty lines inside methods, and please reduce AI usage to the minimum, as the code is bloated due to it.

rrebollo:
@pedrobaeza what do U think? can we merge it now?

stock_picking_report_valued_sale_mrp/tests/test_stock_picking_report_valued_mrp.py
```diff
@@ -73,3 +86,69 @@ def test_01_picking_confirmed(self):
...
+        )
+        return_wiz = return_wizard_form.save()
+        res = return_wiz.create_returns()["res_id"]
+        return_picking = self.env["stock.picking"].browse(res)
+        return_picking.action_assign()
+        for line in return_picking.move_ids:
+            line.quantity_done = line.product_uom_qty
+        return_picking.button_validate()
+        # Redelivery
+        sale._action_cancel()
```
pedrobaeza, line 122: This is not the usual flow, but to return the return, so I'm not doubting your code serves for this purpose.

rrebollo: Do you mean "now doubting"? The first one to question my code was me:

> ### 🛠️ Proposed Fix
> This PR provides an initial attempt to resolve the incorrect quantity computation. Additional details and discussion can be found in the linked issue.

The issue is documented in #451. Please take some time to suggest a better approach if you have one.

rrebollo: @pedrobaeza, now I understand your point. The usual flow is to return the return not to cancel the return. I need to dig deeper into those scenarios. Thank you.

PR #456, opened 2025-11-25
[19.0][MIG] stock_account_valuation_report

AaronHForgeFlow:
The stock value info is taken from the stock move instead of the layers


gdgellatly (approved):
Static review + Functional test on live migrated db with 80k products and 10s of millions of moves.  Essentially functionally equivelant despite move from svl's  It is still really slow but works as expected, maybe marginally better than before.

Update: OK really slow is understating this, before it was 2 or 3 minutes, now still loading after 15 mins when filter set to Valuation Discrepancy.

Alright effectively unusable on large db's although I can see it works.

Final update: It loaded 80 records using Valuation discrepancy after 1 hour. I scrolled to the next 80 in the hope it had them all. It did not. I feel like the idea of this module has value, but we need to rethink how it is done.

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,178 @@
...
+        # pylint: disable=E8103
+        query = """
+            SELECT aml.product_id, aml.account_id,
+                sum(aml.balance),
+                sum(CASE WHEN aml.balance < 0 THEN -aml.quantity ELSE aml.quantity END),
+                array_agg(aml.id)
+            FROM account_move_line AS aml
+            INNER JOIN account_move AS am ON am.id = aml.move_id
+            WHERE aml.product_id IN %s
+            AND am.state = 'posted'
```
gdgellatly: This should use parent_state and avoid join

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+from odoo import api, fields, models
+from odoo.tools import float_compare
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    stock_value = fields.Float("Inventory Value", compute="_compute_inventory_value")
```
gdgellatly: How does this differ from total_value on product_product? Also why not fields.Monetary

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,178 @@
...
+    account_value = fields.Float("Accounting Value", compute="_compute_inventory_value")
+    qty_at_date = fields.Float("Inventory Quantity", compute="_compute_inventory_value")
```
gdgellatly: How does this differ from the standard qty_available and also why not digits?

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,178 @@
...
+    account_qty_at_date = fields.Float(
+        "Accounting Quantity", compute="_compute_inventory_value"
+    )
+    stock_fifo_real_time_aml_ids = fields.Many2many(
```
gdgellatly: I don't think we need to precompute these, the button only does one at a time so we can get them at click for 1 product

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,178 @@
...
+        "account.move.line", compute="_compute_inventory_value"
+    )
+    stock_move_valuated_ids = fields.Many2many(
```
gdgellatly: As above

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,178 @@
...
+        "stock.move", compute="_compute_inventory_value"
+    )
+    valuation_discrepancy = fields.Float(
```
gdgellatly: Also think fields.Monetary then we cna perhaps avoid some of the individual comparison calls

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        products_with_discrepancy = products.filtered(
+            lambda pp: self.env.company.currency_id.compare_amounts(
+                pp.stock_value, pp.account_value
+            )
+        )
+        return [("id", "in", products_with_discrepancy.ids)]
+
+    def _compute_inventory_value(self):
+        self.env["account.move.line"].check_access("read")
```
gdgellatly: here we are only interested in the valuation account but we are fetching for every single account. logically there are only a few valuation accounts at most and nearly 4x as many amls (non AS) and 8x (AS). I think we can greatly cut the return of this query by querying only valuation accounts. I suppose I also have a question of what happens to this report if I change the valuation account or category of a product.

gdgellatly (commented):
I am trying to make this more performant, but honestly, so far not much luck, just what I've found so far.

Also I think we canhandle the zero cases better. If there are no aml's and/or no moves we can skip many computations.

gdgellatly:
I have an alternative implementation on top of this PR that addresses significant performance issues and a bug we encountered:

**Branch:** [`gdgellatly:19.0-gdg-valuation-perf`](https://github.com/gdgellatly/stock-logistics-reporting/tree/19.0-gdg-valuation-perf)

### Bug fix

The existing accounting quantity calculation uses a `CASE WHEN aml.balance < 0 THEN -aml.quantity ELSE aml.quantity END` pattern. This was correct in older Odoo versions where `aml.quantity` was always unsigned (positive), but **in v19 `aml.quantity` is already signed** (negative for credit/outgoing lines). The CASE double-negates outgoing quantities, making them positive — so outgoing goods are *added* to the accounting quantity instead of subtracted.

**Fix:** replaced with `sum(aml.quantity)`.

### Performance changes

- **Use core's `total_value` and `qty_available`** instead of iterating `stock.move` records and summing `remaining_qty`/`remaining_value` — leverages Odoo's batch-optimised computation
- **Pre-filter products via SQL** — only storable products with done stock moves are considered, skipping thousands of irrelevant products
- **Filter AML query by valuation accounts only** — there are typically very few valuation accounts per company, so adding `account_id IN (...)` dramatically reduces the rows scanned
- **Approximate stock value in discrepancy search for AVCO/Standard** — uses `qty_available * standard_price` for Standard and AVCO products instead of triggering the expensive per-move replay during the search filter. FIFO products still use the exact `total_value`. The accurate `total_value` from core is always displayed in the list view for the rendered page.
- **Removed pre-computed Many2many fields** (`stock_fifo_real_time_aml_ids`, `stock_move_valuated_ids`) — button drill-downs now build their domains on click, avoiding bulk computation
- **`fields.Monetary`** with `company_currency_id` for value fields, **`digits="Product Unit"`** for quantity fields
- Aligned wizard context key from `at_date` to core's `to_date`
- Updated view: replaced `stock_value` with core `total_value`, fixed `is_storable` domain, removed redundant `widget="monetary"` attributes

### Performance

On a database with ~2000 storable products and several million stock moves, the valuation discrepancy filter went from **over 60 minutes** (original implementation timed out) to **~30 seconds**. The remaining ~30s is dominated by core's `_compute_value` rendering `total_value` for the displayed page (FIFO/AVCO replay), which is outside the scope of this module.

### Caveat

The valuation discrepancy search uses `qty_available * standard_price` as a proxy for stock value for **Standard and AVCO** cost methods only. This is:
- **Exact** for Standard Cost products
- **Very close** for AVCO (standard_price is updated to weighted average on each move)
- **FIFO products use the exact `total_value`** — no approximation

This means the search filter results should be equivalent for Standard and FIFO, and near-equivalent for AVCO. The accurate value is always shown in the list view itself.

gdgellatly:
OK I give up, this is almost impossible, the quantity field on aml is used differently in different operations, I just can't get it to make any sense. whether I use the current algorithm (definitely wrong), quantity, or -quantity. Anyway hopefully some of this helps

AaronHForgeFlow:
Hi @gdgellatly, thank you, I am afraid I have not had the time to review your branch. 

What's the overall performance improvement for your large database?  I'll review it and likely merge it to this branch.

gdgellatly:
> Hi @gdgellatly, thank you, I am afraid I have not had the time to review your branch.
> 
> What's the overall performance improvement for your large database? I'll review it and likely merge it to this branch.

@AaronHForgeFlow It is just a bunch of experiments, and not very well working ones at that. I would not merge it. I actually feel like we might need to go back to the old v7 days where we used to maintain historical aggregates to lower the search window. I will relook later this week to see if we can more easily integrate that approach because it won't be just this module, just this module because of it's non stored computes is super badly affected

AaronHForgeFlow:
Thank you @gdgellatly , I applied the code fixes suggestions. Also partial of the performance improvements suggestions. In a small size database (>10000 products, around (1000 stock moves and journal items)  the performance improvement is significant (from 5 seconds to almost immediate). Likely not sufficient for large databases, but I think they are useful for small companies.

For large companies, we can store the report and run it overnight with an scheduled action, but I prefer to wait until I have a real use case in v19.

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
+# Copyright 2020 ForgeFlow S.L.
+# Copyright 2019 Aleph Objects, Inc.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import api, fields, models
```
rrebollo, line 6:
```suggestion
from odoo import api, fields, models
from odoo.fields import Domain
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+    )
+    qty_discrepancy = fields.Float(
+        compute="_compute_inventory_value",
+        search="_search_qty_discrepancy",
+        digits="Product Unit of Measure",
+    )
+
+    @api.model
+    def _search_qty_discrepancy(self, operator, value):
+        return [("id", "in", self._get_discrepancy_product_ids("qty"))]
```
rrebollo:
```suggestion
        return Domain("id", "in", self._get_discrepancy_product_ids("qty"))
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+
+    @api.model
+    def _search_valuation_discrepancy(self, operator, value):
+        return [("id", "in", self._get_discrepancy_product_ids("value"))]
```
rrebollo:
```suggestion
        return Domain("id", "in", self._get_discrepancy_product_ids("value"))
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+
+    @api.model
+    def _get_discrepancy_product_ids(self, kind):
+        """Return product IDs where stock and accounting values/qtys diverge."""
+        to_date = self.env.context.get("at_date", False)
+        company_id = self.env.company.id
+        categories = self.env["product.category"].search(
+            [("property_valuation", "=", "real_time")]
```
rrebollo:
```suggestion
            Domain("property_valuation", "=", "real_time")
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+            precision_kwargs = {
+                "precision_digits": self.env["decimal.precision"].precision_get(
+                    "Product Unit of Measure"
+                )
+            }
+        move_domain = [
+            ("product_id", "in", product_ids),
+            ("company_id", "=", company_id),
+            ("state", "=", "done"),
+        ]
```
rrebollo:
```suggestion
        move_domain = Domain("product_id", "in", product_ids) &
            Domain("company_id", "=", company_id) &
            Domain("state", "=", "done")
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+        if to_date:
+            move_domain.append(("date", "<=", to_date))
```
rrebollo:
```suggestion
            move_domain = move_domian & Domain("date", "<=", to_date)
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+                params = params + (to_date,)
+            self.env.cr.execute(query, params=params)
+            for row in self.env.cr.fetchall():
+                accounting_values[(row[0], row[1])] = (row[2], row[3])
+        # 2) INVENTORY VALUES
+        move_domain = [
+            ("product_id", "in", self.ids),
+            ("company_id", "=", self.env.company.id),
+            ("state", "=", "done"),
+        ]
```
rrebollo:
```suggestion
        move_domain = Domain("product_id", "in", self.ids) &
            Domain("company_id", "=", self.env.company.id) &
            Domain("state", "=", "done")
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+        if to_date:
+            move_domain.append(("date", "<=", to_date))
```
rrebollo:
```suggestion
            move_domain = move_domain & Domain("date", "<=", to_date)
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+        list_view_ref = self.env.ref("account.view_move_line_tree")
+        form_view_ref = self.env.ref("account.view_move_line_form")
+        return {
+            "name": self.env._("Accounting Valuation at date"),
+            "type": "ir.actions.act_window",
+            "view_type": "form",
+            "view_mode": "list,form",
+            "context": self.env.context,
+            "res_model": "account.move.line",
+            "domain": [("id", "in", aml_ids)],
```
rrebollo:
```suggestion
            "domain": Domain("id", "in", aml_ids),
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+        self.env.cr.execute(query, params=params)
+        return [row[0] for row in self.env.cr.fetchall()]
+
+    def action_view_valuation_layers(self):
+        to_date = self.env.context.get("at_date", False)
+        move_ids = self._get_move_ids(self.ids, to_date)
+        action = self.env["ir.actions.actions"]._for_xml_id(
+            "stock_account.stock_valuation_layer_report_action"
+        )
+        action["domain"] = [("id", "in", move_ids)]
```
rrebollo:
```suggestion
        action["domain"] = Domain("id", "in", move_ids)
```

stock_account_valuation_report/models/product_product.py
```diff
@@ -0,0 +1,301 @@
...
+        action["context"] = {}
+        return action
+
+    def action_view_valuation_moves(self):
+        self.ensure_one()
+        to_date = self.env.context.get("at_date", False)
+        move_ids = self._get_move_ids(self.ids, to_date)
+        action = self.env["ir.actions.actions"]._for_xml_id("stock.stock_move_action")
+        action["domain"] = [("id", "in", move_ids)]
```
rrebollo:
```suggestion
        action["domain"] = Domain("id", "in", move_ids)
```

stock_account_valuation_report/tests/test_stock_account_valuation_report.py
```diff
@@ -0,0 +1,119 @@
+# 2020 Copyright ForgeFlow, S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+
+from odoo.addons.stock_account.tests.common import TestStockValuationCommon
```
rrebollo, line 7:
```suggestion
from odoo.addons.stock_account.tests.common import TestStockValuationCommon
from odoo.fields import Domain
```

stock_account_valuation_report/tests/test_stock_account_valuation_report.py
```diff
@@ -0,0 +1,119 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.aml_model = cls.env["account.move.line"]
+
+    def test_01_stock_receipt(self):
+        """Receive into stock and ship to the customer"""
+        self._use_inventory_location_accounting()
+        move = self._make_in_move(self.product_fifo_auto, quantity=1, unit_cost=10.0)
+        inv_aml = self.aml_model.search(
+            [("product_id", "=", self.product_fifo_auto.id)]
```
rrebollo:
```suggestion
            Domain("product_id", "=", self.product_fifo_auto.id)
```

stock_account_valuation_report/tests/test_stock_account_valuation_report.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertEqual(self.product_fifo_auto.qty_at_date, 0.0)
+        self.assertEqual(self.product_fifo_auto.account_qty_at_date, 0.0)
+
+    def test_02_stock_receipt_several_costs_several_dates(self):
+        """Receive into stock at different cost"""
+        self._use_inventory_location_accounting()
+        move = self._make_in_move(self.product_fifo_auto, quantity=1.0, unit_cost=10.0)
+        # This will not create any journal entry
+        inv_aml = self.aml_model.search(
+            [("product_id", "=", self.product_fifo_auto.id)]
```
rrebollo:
```suggestion
            Domain("product_id", "=", self.product_fifo_auto.id)
```

stock_account_valuation_report/tests/test_stock_account_valuation_report.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        balance_inv = sum(inv_aml.mapped("balance"))
+        self.assertEqual(balance_inv, 0.0)
+        self.assertEqual(move.remaining_value, 10.0)
+        # Receive more
+        move2 = self._make_in_move(self.product_fifo_auto, quantity=2.0, unit_cost=20.0)
+        # This will not create any journal entry
+        inv_aml = self.aml_model.search(
+            [("product_id", "=", self.product_fifo_auto.id)]
```
rrebollo:
```suggestion
            Domain("product_id", "=", self.product_fifo_auto.id)
```

stock_account_valuation_report/tests/test_stock_account_valuation_report.py
```diff
@@ -0,0 +1,119 @@
...
+        self.product_fifo_auto._compute_inventory_value()
+        self.assertEqual(self.product_fifo_auto.stock_value, 10.0)
+        self.assertEqual(self.product_fifo_auto.account_value, 10.0)
+        self.assertEqual(self.product_fifo_auto.qty_at_date, 1.0)
+        self.assertEqual(self.product_fifo_auto.account_qty_at_date, 1.0)
+        self.assertEqual(self.product_fifo_auto.valuation_discrepancy, 0.0)
+        self.assertEqual(self.product_fifo_auto.qty_discrepancy, 0.0)
+        # check accounting entries
+        inv_aml = self.aml_model.search(
+            [("product_id", "=", self.product_fifo_auto.id)]
```
rrebollo:
```suggestion
            Domain("product_id", "=", self.product_fifo_auto.id)
```

stock_account_valuation_report/wizards/stock_valuation_history.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def open_at_date(self):
+        action = self.env["ir.actions.act_window"]._for_xml_id(
+            "stock_account_valuation_report.product_valuation_action"
+        )
+        domain = Domain("is_storable", "=", True)
+        product_id = self.env.context.get("product_id", False)
+        product_tmpl_id = self.env.context.get("product_tmpl_id", False)
+        if product_id:
+            domain = Domain.AND([domain, [("id", "=", product_id)]])
```
rrebollo:
```suggestion
            domain = domain & Domain("id", "=", product_id)
```

stock_account_valuation_report/wizards/stock_valuation_history.py
```diff
@@ -0,0 +1,35 @@
...
+        elif product_tmpl_id:
+            domain = Domain.AND([domain, [("product_tmpl_id", "=", product_tmpl_id)]])
```
rrebollo:
```suggestion
            domain = domain & Domain("product_tmpl_id", "=", product_tmpl_id)
```

rrebollo (commented):
I did some suggestions for your consideration

rrebollo:
@AaronHForgeFlow would you be so kind to review my #479 in return?

AaronHForgeFlow:
@rrebollo Thank you for the suggestions, I applied them all.

rrebollo (approved):
Code Review. LGTM!

BhaveshHeliconia (approved):
Functional review LGTM!

gdgellatly:
> Thank you @gdgellatly , I applied the code fixes suggestions. Also partial of the performance improvements suggestions. In a small size database (>10000 products, around (1000 stock moves and journal items) the performance improvement is significant (from 5 seconds to almost immediate). Likely not sufficient for large databases, but I think they are useful for small companies.
> 
> For large companies, we can store the report and run it overnight with an scheduled action, but I prefer to wait until I have a real use case in v19.

Hi @AaronHForgeFlow I have not got back to this. But my current thinking is that we create markers to reduce the moves under consideration. You will notice that in certain operation such as changing standard price odoo creates a stock value record. When computing stock value it limits the moves search to the earliest of these records for AVCO/FIFO replay. I have not looked at how to implement yet.

We have also traced back most of the performance issues to lot valuated products, so the scope of issues may be limited and I am working with Odoo to see what we can do.

stock_account_valuation_report/views/product_product_views.xml
```diff
@@ -0,0 +1,108 @@
...
+            </list>
+        </field>
+    </record>
+
+    <record id="product_valuation_action" model="ir.actions.act_window">
+        <field name="name">Dual Inventory Valuation</field>
+        <field name="res_model">product.product</field>
+        <field name="view_mode">list,form</field>
+        <field name="view_id" ref="view_stock_product_tree2" />
+        <field name="domain">[('type', '=', 'product')]</field>
```
dnplkndll:
```suggestion
          <field name="domain">[('is_storable', '=', True)]</field>
  ```

File: stock_account_valuation_report/models/product_product.py [hunk 27]
```diff
@@ -0,0 +1,302 @@
...
+
+        accounting_values = {}
+        if valuation_account_ids:
+            self.env["account.move.line"].flush_model()
+            # pylint: disable=E8103
+            query = """
+                SELECT aml.product_id, aml.account_id,
+                    sum(aml.balance),
+                    sum(CASE WHEN aml.balance < 0
+                        THEN -aml.quantity ELSE aml.quantity END)
```
dnplkndll: @AaronHForgeFlow stock_account migration ([commit ](https://github.com/OCA/OpenUpgrade/commit/9d927ccaf342996374e7456f484ff380cc59725a)[9d927cc](https://github.com/OCA/OpenUpgrade/commit/9d927ccaf342996374e7456f484ff380cc59725a)) relinks stock.move ↔ account.move ([post-migration L22](https://github.com/OCA/OpenUpgrade/blob/9d927ccaf342996374e7456f484ff380cc59725a/openupgrade_scripts/scripts/stock_account/19.0.1.1/post-migration.py#L22)) but never rewrites account_move_line.quantity. So migrated DBs keep the legacy signed valuation-line quantities, whereas native 19 produces unsigned (COGS) + zero (stock-move) — which is what the CASE WHEN aml.balance < 0 … report SQL assumes. 

Is preserving the legacy quantity the intended migration shape, or should the stock-move valuation lines be normalized to the 19.0 zero-quantity form?

dnplkndll (changes requested):

stock_account_valuation_report/models/product_product.py
AaronHForgeFlow: Good point. Removed the quantity comparison completely, it is not possible in v19.

AaronHForgeFlow: Actually, I think that the most common case is still covered. When you get the vendor bill the qty is increased and when you receive the customer invoice the qty is decreased.

For the inventory adjustments or scrap operations the qty_discrepancy won't work and that is stated in the ROADMAP.

In regards legacy data, I think it should not be covered in the module. That should be discussed as part of the stock_account migration scripts.

dnplkndll (approved):

PR #457, opened 2025-12-16
[16.0][IMP] stock_picking_report_valued: Add conditional Total Picking block in valued delivery slip report

Anxo82:
@HaraldPanten @Jaimermaccione @ValentinVinagre 

- Add a conditional “Total Picking” block to the valued delivery slip report.  An extensible hook is introduced to allow other modules to adjust the total amount printed in the report (e.g. taxes, fees or surcharges), while keeping the default behavior unchanged when no adjustment is applied.

[T-9158]


Jaimermaccione (approved):
@Anxo82  LGTM! The module has been reviewed in a local environment. All functional tests performed behave as expected.

#### Tests performed

1. **Delivery slip printing without any applied tax**  
   - The report is not affected by the module.  
   - The table is displayed exactly as in the previous version, without adding the picking total.

2. **Delivery slip printing with a tax breakdown module enabled**  
   - When used together with a module that adds a tax breakdown to the delivery slip, the report is generated correctly.  
   - The **picking total** is displayed correctly, including the applied tax.

#### Conclusion

Functional validation is successful. The module does not introduce side effects and behaves correctly both in scenarios without taxes and in scenarios where a tax breakdown is applied.

ValentinVinagre:
@Anxo82 Could you provide an example of what a report looks like with this change?
Thanks

Anxo82:
[Delivery Slip - Deco Addict - WH_OUT_00021.pdf](https://github.com/user-attachments/files/24327040/Delivery.Slip.-.Deco.Addict.-.WH_OUT_00021.pdf)
Screenshot showing the addition of the total block in the view:

image

ValentinVinagre (approved):
LGTM 👍🏻

Anxo82:
@HaraldPanten has already been reviewed. Could you request the merge?
Thanks

HaraldPanten (approved):
@pedrobaeza @etobella Could you merge this PR? Let me explain why this change is needed, giving you more context:

In Spain, we have modules such as l10n_es_sigaus_stock_picking_report_valued that display the total amount in valued pickings.

In the future, we will also have other modules that handle SCRAP fees (we're working on that).

What happens when a customer works with several different SCRAP fees in their operations?

In that case, the customer needs to unify the totals shown in the valued picking.

With this change, the current module continues to work exactly as before, but it is now prepared to support the scenarios described above.

Thanks!

stock_picking_report_valued/report/stock_picking_report_valued.xml
```diff
@@ -85,6 +85,40 @@
...
+                    t-if="o._show_report_valued_total_block()"
+                >
+                    <table
+                        class="table table-sm"
+                        name="valued_picking_total_table"
+                        style="background-color: rgba(0, 0, 0, 0.1);"
+                    >
+                        <tr>
+                            <td>
+                                <strong>Total Picking</strong>
```
pedrobaeza: This label doesn't seem correct according what you are mentioning. Shouldn't it be `Total Block`?

Anxo82: Hi @pedrobaeza , I think “Total Picking” makes more sense here, because this block is meant to display the final valued total of the delivery slip, including any additional amounts added by other modules (such as SIGAUS,  or other future fees/valuations):

image

With this approach, we can also remove the custom “total” block currently implemented in the SIGAUS module 
https://github.com/OCA/l10n-spain/blob/cb464963d188c069c266c54ba6a85309e54870c2/l10n_es_sigaus_stock_picking_report_valued/report/report_deliveryslip.xml#L43, since this common block becomes the single extension point for adjusting the printed picking total.
This avoids duplication and means that other modules won’t need to implement their own total blocks anymore. Thanks

pedrobaeza: Well, I think we can remove at least the word "Picking". Let's keep just "Total".

Anxo82: Done, thanks.

etobella (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #466, opened 2026-01-13
[16.0][FIX] stock_picking_report_valued: handle missing currency on manual move lines

atgalvez08:
@BinhexTeam

Problem

When using the module stock_picking_report_valued, printing the valued picking report may fail with:
ValueError: Expected singleton: res.currency()

This happens when a user manually adds stock move lines to a picking that is already done.
These move lines are not linked to a sale order line and therefore do not have a currency set.

Since the report displays sale_price_subtotal as a Monetary field, the QWeb renderer expects a valid currency and crashes when none is found.

This scenario is fully allowed by Odoo and represents a valid user workflow.

Root Cause

The valued picking report assumes that all stock move lines have an associated currency coming from the sale order.

However, manually added move lines:
	•	are not linked to a sale order line
	•	do not have currency_id set
	•	still expose sale_price_subtotal as a Monetary field

This results in a QWeb rendering error.

Solution

Make the report rendering more robust by explicitly defining a fallback currency at QWeb level.

The currency is resolved in the following order:
	1.	move_line.currency_id
	2.	move_line.move_id.sale_line_id.currency_id
	3.	move_line.company_id.currency_id

This guarantees that:
	•	the report never crashes
	•	values remain unchanged
	•	the fix is limited to rendering logic only

Notes on Testing

This change only affects QWeb report rendering and does not modify business logic or data.

Given the scope and the complexity of testing QWeb rendering in CI, no automated test has been added, following existing precedent in this repository.

### DIFF
--- a/stock_picking_report_valued/report/stock_picking_report_valued.xml
+++ b/stock_picking_report_valued/report/stock_picking_report_valued.xml
@@ -7,8 +7,10 @@
         </xpath>
         <xpath expr="//th[@name='th_sml_quantity']" position="after">
             <t t-if="o.valued and o.sale_id and o.move_line_ids and is_outgoing">
-                <th name="th_sml_unit_price" class="text-end"><strong
-                    >Unit Price</strong></th>
+                <th name="th_sml_unit_price" class="text-end">
+                    <strong>Unit Price
+                    </strong>
+                </th>
                 <th
                     name="th_sml_discount"
                     class="text-end"
@@ -16,9 +18,13 @@
                 >
                     <strong>Discount</strong>
                 </th>
-                <th name="th_sml_subtotal" class="text-end"><strong
-                    >Subtotal</strong></th>
-                <th name="th_sml_taxes" class="text-end"><strong>Taxes</strong></th>
+                <th name="th_sml_subtotal" class="text-end">
+                    <strong>Subtotal
+                    </strong>
+                </th>
+                <th name="th_sml_taxes" class="text-end">
+                    <strong>Taxes</strong>
+                </th>
             </t>
         </xpath>
         <xpath expr="//th[@name='th_sml_qty_ordered']" position="attributes">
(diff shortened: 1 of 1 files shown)


adasatorres (approved):
Code review, LGTM.

jelenapoblet:
I did a functional review on this, but I think something it's not working correctly.
This is the flow tested:
I selected a Partner with Valued Picking enabled -> created a SO-> confirmed-> print delivery slip -> OK 
image
I added more products on the picking level -> printed delivery slip -> ammount added on the report -> Not valuated in the total.
image
The product has a sales price, so I think adding this price on the total picking section shoud be added.

atgalvez08:
> I did a functional review on this, but I think something it's not working correctly. This is the flow tested: I selected a Partner with Valued Picking enabled -> created a SO-> confirmed-> print delivery slip -> OK image I added more products on the picking level -> printed delivery slip -> ammount added on the report -> Not valuated in the total. image The product has a sales price, so I think adding this price on the total picking section shoud be added.


Thanks for the feedback.

The original crash happens because `sale_price_subtotal` is a Monetary field
that depends on `currency_id`, which is only defined when the move line is
linked to a sale order line. When users manually add move lines at picking
level, no `sale_line_id` nor currency exists, causing the report to crash.

This PR addresses that issue by making the report rendering defensive and
ensuring a valid display currency is always provided, preventing the QWeb
Monetary widget from failing.

Regarding totals: manually added move lines are not linked to a sale order
line and therefore do not participate in the commercial valuation logic
implemented in `_compute_sale_order_line_fields`. Including such lines in
the total would require defining new business rules (which price, taxes,
currency should apply) and is outside the scope of this fix.

This PR intentionally focuses on restoring report stability without changing
valuation semantics. A functional enhancement to include manually added lines
in totals could be discussed in a separate PR.

jelenapoblet (approved):
Ok then, this fix resolves the case you're mentioning.

sheragar (approved):
LGTM. Funtional Review 👍

pedrobaeza:
@jelenapoblet next time, for fixes, use patch or minor, not nobump.

PR #467, opened 2026-01-14
[16.0][FIX] stock_picking_report_valued: Fix rounding of valued picking total

Anxo82:
@HaraldPanten @Jaimermaccione

- This PR fixes rounding issues in the valued delivery slip total when additional amounts are added. The issue could appear on some valued pickings due to float precision, but not consistently on all of them.

Before:
image
After:
image

- [I-8204]


stock_picking_report_valued/models/stock_picking.py
```diff
@@ -66,6 +66,11 @@ def _get_report_valued_total_amount(self):
         self.ensure_one()
         return self.amount_total
 
+    def _get_report_valued_total_amount_round(self):
```
ValentinVinagre: Change the comment in the function "_get_report_valued_total_amount" and update the readme file. Thanks.

ValentinVinagre (changes requested):

Jaimermaccione (approved):
@Anxo82 LGTM! Review completed in the local environment: total decimals are displayed correctly in the picking PDF report.

stock_picking_report_valued/models/stock_picking.py
Anxo82: Done

ValentinVinagre (changes requested):
add tests please

ValentinVinagre (approved):
LGTM 👍🏻

Anxo82:
@JordiBForgeFlow  or @rousseldenis , could you please review this PR for merge when you have a moment?
Thanks!

Jaimermaccione:
@JordiBForgeFlow @rousseldenis Just a gentle reminder to take a look at this PR when you have a moment, in case it can be merged.
Thanks!

HaraldPanten (approved):
@etobella @pedrobaeza Could you merge this FIX, please?

THX!

pedrobaeza (changes requested):
You are proposing a workaround for another problem, which is that for that picking, the currency_id is not set, and that's why the round is not done. Please fix the underlying problem.

Anxo82:
Hi @pedrobaeza sorry for the delayed response — I’ve been away due to personal reasons.

Thanks for the feedback.

After reviewing this in detail with logs, I found that the issue was not caused by a missing `currency_id`, but by how the total was rendered in the report.

The method `_get_report_valued_total_amount()` returns a Python float, which in some cases led to floating-point precision artifacts (e.g. 569.689999999999) being displayed.

The root cause was that the value was rendered using `t-esc` with the legacy widget="monetary" syntax, which does not properly format the result of a method call.

This has been fixed by switching to the proper `t-out + t-options` syntax for the monetary widget, so the value is now correctly formatted.

Additionally, I verified that even in cases where currency_id is not explicitly set on the picking, the company currency is used as a fallback, so the rounding behavior remains correct.

Therefore, this change fixes the underlying rendering issue rather than introducing a workaround.

Thanks

pedrobaeza (approved):
I still think the problem is `currency_id` is not set, so your workaround is to fallback to company currency with the new syntax, but this may be reflected in another field, but let's unblock this.

/ocabot merge patch

PR #473, opened 2026-02-19
[18.0][FIX] stock_account_valuation_report: avoid field override

LauraCForgeFlow:
The `stock_valuation_layer_ids` field was overriding a standard Odoo field (see [original definition](https://github.com/odoo/odoo/blob/fd42e8313bd70269d1150fb1dd162e56561a2bd8/addons/stock_account/models/product.py#L148)), changing its type, the store condition, and even the inverse relationship with the `stock.valuation.layer` model through the `product_id` field. 

These caused issues with computed field cache invalidation. Specifically, creating or writing new stock valuation layers no longer triggered automatic recomputation of dependent product fields such as `value_svl` and `quantity_svl`.

To fix this, we renamed the field instead of overriding the standard field. This avoids conflicts with dependency tracking and ensures cache invalidation works correctly. Because the field is originally computed and non-stored, there is no need to implement a migration script.

### DIFF
--- a/stock_account_valuation_report/models/product_product.py
+++ b/stock_account_valuation_report/models/product_product.py
@@ -19,7 +19,7 @@ class ProductProduct(models.Model):
     stock_fifo_real_time_aml_ids = fields.Many2many(
         "account.move.line", compute="_compute_inventory_value"
     )
-    stock_valuation_layer_ids = fields.Many2many(
+    valuation_layer_ids = fields.Many2many(
         "stock.valuation.layer", compute="_compute_inventory_value"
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
Thanks a lot for the fix! 🙌 

IMHO this should be merged ASAP, overwriting a standard field that way was never the purpose of this module. The change in the name does not affect the current usability of the module because it is just a computed field.

LoisRForgeFlow (approved):
Ty! 

/ocabot merge patch

PR #474, opened 2026-02-19
[18.0][FIX] stock_picking_report_valued: method patching in tests

LauraCForgeFlow:
Odoo implemented this [commit](https://github.com/odoo/odoo/commit/de056cc784a3bbe2575fd3c9e81ca62e73c362d4) which restricts and checks for methods being set or patched inside tests, so we needed to adapt to that.


LoisRForgeFlow (approved):
Thanks!

PR #478, opened 2026-02-26
[15.0][FIX] stock_picking_report_undelivered_product: Error when trying to sort products with and without default_code

CarlosRoca13:
When a picking has moves with products that has default_code and other moves that hasn't it. The system throws an error when trying to print the Delivery Note and they are not delivered yet.

cc @Tecnativa TT61225

ping @sergio-teruel @carlosdauden


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #479, opened 2026-02-28
[17.0] [MIG] stock_delivery_note

rrebollo:
Standard migration, backport from v18. Bot commits has been squashed.

#297
@BinhexTeam T19998

PR #481, opened 2026-03-12
[18.0][MIG] stock_average_daily_sale

rousseldenis:


rousseldenis:
@jbaudoux

jbaudoux (commented):
can you drop the migrations folder?

rousseldenis:
@jbaudoux ready

jbaudoux (approved):

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -0,0 +1,354 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+from contextlib import closing
+from math import sqrt
+
+from psycopg2.errors import ObjectNotInPrerequisiteState
+from psycopg2.extensions import AsIs
+
+from odoo import _, api, fields, models, registry
```
rrebollo, line 13:
```suggestion
from odoo import api, fields, models
from odoo.modules.registry import Registry
```
From OCA Migration Guidelines to v18:

>Instead of doing from odoo import registry and then registry(db_name), now you have to do from odoo.modules.registry import Registry and Registry(db_name). More info at https://github.com/odoo/odoo/pull/178784.

Also translations are environment related now.

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -0,0 +1,354 @@
...
+        cr.execute(
+            "SELECT ispopulated FROM pg_matviews WHERE matviewname = %s;",
+            (cls._table,),
+        )
+        records = cr.fetchone()
+        return records and records[0]
+
+    @api.model
+    def _check_view(self):
+        cr = registry(self._cr.dbname).cursor()
```
rrebollo, line 134:
```suggestion
        cr = Registry(self._cr.dbname).cursor()
```
Idem

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -0,0 +1,354 @@
...
+                return False
+            except Exception as e:
+                raise e
+
+    # pylint: disable=redefined-outer-name
+    @api.model
+    def search(self, domain, offset=0, limit=None, order=None):
+        if not config["test_enable"] and not self._check_view():
+            return self.browse()
+        return super().search(domain=domain, offset=offset, limit=limit, order=order)
```
rrebollo, line 151: Did you consider OCA Migration Guidelines? Given this model is really a materialized table I'm guessing this is not relevant. Am I wrong?

>If overriding search method for modifying searches result, take into account that this method is not always called in this version. Override search_fetch instead.

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -0,0 +1,85 @@
...
+
+from odoo.addons.stock_storage_type_putaway_abc.models.stock_location import (
+    ABC_SELECTION,
+)
+
+
+class StockAverageDailySaleConfig(models.Model):
+    _name = "stock.average.daily.sale.config"
+    _description = "Average daily sales computation parameters"
+    check_company_auto = True
```
rrebollo, line 15:
```suggestion
    _check_company_auto = True
```
Looks like a typo. Anyway I think it's better to address this now.

rrebollo (changes requested):
I believe most of the findings are optional, but a few need to be addressed at this point. For all those cases, I've provided a suggestion that should make the change straightforward.

rrebollo:
Also @rousseldenis, would you be so kind to review my #479 in return?

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @api.model
+    def _check_view(self):
+        cr = registry(self._cr.dbname).cursor()
+        with closing(cr):
+            try:
+                return self._check_materialize_view_populated(cr)
+            except ObjectNotInPrerequisiteState:
+                _logger.warning(
+                    _("The materialized view has not been populated. Launch the cron.")
```
rrebollo, line 140:
```suggestion
                    self.env._("The materialized view has not been populated. Launch the cron.")
```

stock_average_daily_sale/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Average Daily Sale",
```
jbaudoux, line 5: Can we rename the module to
```suggestion
    "name": "Stock Average Daily Usage",
```

stock_average_daily_sale/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "summary": """
+        Allows to gather consumed products on daily basis""",
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,BCIM,Odoo Community Association (OCA)",
+    "maintainers": ["jbaudoux"],
+    "website": "https://github.com/OCA/stock-logistics-reporting",
+    "depends": [
+        "sale",
```
jbaudoux, line 14: It should only depend on stock

stock_average_daily_sale/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+        "stock_storage_type_putaway_abc",
```
jbaudoux, line 15: I would like to drop this strong dependency and move it to a glue module. Or even better, replace it with a products applicability domain. At the end, it is the only thing that it is doing.

Also it would be better to use the abc product classification level that is at product.product level (abc_storage is only at product.template level) and that supports multi-warehouses through the abc profiles.

stock_average_daily_sale/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+        "product_route_mto",
```
jbaudoux, line 16: MTO should move to another glue module. It doesn't provide any value to this ADU computation.

New module: stock_average_daily_usage_mto

stock_average_daily_sale/models/stock_average_daily_sale.py
```diff
@@ -0,0 +1,354 @@
...
+    date_from = fields.Date(string="From", required=True)
+    date_to = fields.Date(string="To", required=True)
+    is_mto = fields.Boolean(
+        string="On Order",
+        readonly=True,
+        store=True,
+        index=True,
+    )
+    nbr_sales = fields.Integer(
+        string="Number of Sales",
```
jbaudoux, line 68: rename to reflect what it is

stock_average_daily_sale/models/stock_average_daily_sale_config.py
```diff
@@ -0,0 +1,85 @@
...
+
+    active = fields.Boolean(default=True)
+    abc_classification_level = fields.Selection(
+        selection=ABC_SELECTION, required=True, default="b"
+    )
```
jbaudoux, line 20: move out of this module

jbaudoux (commented):
I'm planning to improve and refactor this module in v18.0.
I noted as comment what I intend to change

jbaudoux:
Fix PRs to include:
* https://github.com/OCA/stock-logistics-reporting/pull/499
* https://github.com/OCA/stock-logistics-reporting/pull/500

PR #492, opened 2026-04-14
[17.0][ADD] stock_report_partner_ref: add customer reference to picking reports

atgalvez08:
@BinhexTeam

This module extends stock picking reports to display the customer reference (`partner.ref`).

## Motivation

In many logistics workflows, the partner reference is a key identifier used by customers, suppliers, and external systems (e.g. EDI or retail platforms).

Standard Odoo stock reports (delivery slip and picking operations) do not include this reference, which can complicate customer identification and require manual cross-checking during warehouse operations.

## Features

- Adds the customer reference to:
  - Delivery Slip (`stock.report_delivery_document`)
  - Picking Operations (`stock.report_picking`)
- Uses a simple fallback logic:
  - `picking.partner_id.ref`
  - `picking.partner_id.commercial_partner_id.ref`

## Design

- No changes to data models
- No impact on existing workflows
- Lightweight QWeb inheritance
- Logic kept intentionally minimal and aligned with standard Odoo behavior

## Use cases

- EDI integrations requiring customer codes
- Retail and distribution environments
- Warehouses where customer reference is needed in printed documents

## Future improvements

- Optional configuration to enable/disable display
- Extension to additional stock reports


stock_report_partner_ref/tests/test_stock_report_partner_ref.py
```diff
@@ -0,0 +1,161 @@
+from odoo.tests import TransactionCase
+
+
+class TestStockReportPartnerRef(TransactionCase):
```
rrebollo: You could use `BaseCommon` instead of `TransactionCase`

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

I think having the option to enable or disable this behavior at the `stock.picking.type` level would be a plus.

stock_report_partner_ref/tests/test_stock_report_partner_ref.py
atgalvez08: Thanks for the review and suggestions!

I've updated the tests to use BaseCommon as suggested.

Regarding the configurability at stock.picking.type level, I agree it would be a valuable enhancement. It is already aligned with the roadmap as optional configuration for the reference display, but I would prefer to handle it in a separate PR to keep this one focused and minimal.

PR #494, opened 2026-05-04
[17.0][FIX] stock_picking_report_custom_description: avoid duplicated references

pilarvargas-tecnativa:
Avoid printing the product when the picking description already contains it.

Some modules or manual edits only change the reference, not the product name, causing duplicated information in delivery slips.

@Tecnativa TT61271

Before:
image

After:
image

@pedrobaeza @carlos-lopez-tecnativa please review

### DIFF
--- a/stock_picking_report_custom_description/report/report_deliveryslip.xml
+++ b/stock_picking_report_custom_description/report/report_deliveryslip.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="report_delivery_document" inherit_id="stock.report_delivery_document">
+        <!-- Avoid duplicated product references when the picking description already
+        includes the product name. -->
+        <xpath expr="//span[@t-field='move.product_id']/.." position="attributes">
+            <attribute name="class">d-none</attribute>
+        </xpath>
+        <xpath expr="//span[@t-field='move.product_id']/.." position="after">
+            <td>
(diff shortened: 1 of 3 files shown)


stock_picking_report_custom_description/report/report_deliveryslip.xml
```diff
@@ -0,0 +1,84 @@
...
+        includes the product name. -->
+        <xpath expr="//span[@t-field='move.product_id']/.." position="attributes">
+            <attribute name="class">d-none</attribute>
+        </xpath>
+        <xpath expr="//span[@t-field='move.product_id']/.." position="after">
+            <td>
+                <t
+                    t-if="not move.description_picking or move.product_id.name not in move.description_picking"
+                >
+                    <span t-field="move.product_id">Customizable Desk</span>
```
pedrobaeza: Do we need here to put a placeholder text?

pedrobaeza (approved):

stock_picking_report_custom_description/report/report_deliveryslip.xml
pilarvargas-tecnativa: No, I don't think so. Since we are simulating the replace, I copied the original node including its placeholder, but it is not needed because the value is rendered by t-field. I’ll remove it to keep the inherit cleaner.

carlos-lopez-tecnativa (approved):

PR #495, opened 2026-05-06
[17.0][FIX] stock_picking_report_custom_description: always transfer sale line description

pilarvargas-tecnativa:
The module is intended to transfer the sale order line description to the picking description.

However, when the sale line description matched the product display name, the description_picking value was not updated, because the assignment was skipped by the condition checking that both values were different.

As a result, the stock move kept the default description generated by Odoo, which could omit the product reference.

Keep the existing logic that trims the product display name prefix when an additional sale description is present, but always assign the resulting description to the picking.

@Tecnativa TT61271

@pedrobaeza @victoralmau please review


pedrobaeza (approved):

victoralmau (approved):
IMO, that makes sense.

Issue #501, opened 2026-06-02
Field no longer exists in Odoo


tobiaszehntner:
## Module
stock_picking_report_valued

## Describe the bug
When no quantity on a move line, we have a stacktrace because it tries to get `reserved_qty` which no longer exists in v19

**Affected versions**: 19.0

**Additional context**
```
  File "/odoo/lib/python3.13/site-packages/odoo/addons/stock_picking_report_valued/models/stock_move_line.py", line 46, in _get_report_valued_quantity
    return self.quantity or self.reserved_qty
                            ^^^^^^^^^^^^^^^^^
AttributeError: 'stock.move.line' object has no attribute 'reserved_qty'
```

PR #502, opened 2026-06-02
[19.0][FIX] stock_picking_report_valued: remove old field use

tobiaszehntner:
`reserved_qty` does no longer exist in v19. `quantity` is now a computed field that shows either reserved or validated quantity.

Fixes https://github.com/OCA/stock-logistics-reporting/issues/501

### DIFF
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -43,7 +43,7 @@ class StockMoveLine(models.Model):
     )
 
     def _get_report_valued_quantity(self):
-        return self.quantity or self.reserved_qty
+        return self.quantity
 
     def _compute_sale_order_line_fields(self):
         """This is computed with sudo for avoiding problems if you don't have


stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -42,9 +42,6 @@ class StockMoveLine(models.Model):
         compute="_compute_sale_order_line_fields", string="Total", compute_sudo=True
     )
 
-    def _get_report_valued_quantity(self):
-        return self.quantity or self.reserved_qty
```
acsonefho, line 46: The best is to keep that function and only remove the `or self.reserved_qty` part.
That keep the compatibility with custom dev and let an inherit point if someone want to use another field value.

acsonefho (changes requested):

acsonefho (approved):

BhaveshHeliconia (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #505, opened 2026-06-11
[FIX] stock_picking_report_valued: set valued_picking default to False

danielamoresdev:
Partners should not have valued pickings enabled by default when the module is installed, as this causes delivery slips to print with prices visible without any explicit configuration.


soniak05 (approved):
LGTM great job!

AdrianaSaiz (approved):
LGTM

BhaveshHeliconia (approved):

PR #510, opened 2026-06-19
[MIG] stock_picking_operations_multilang: Migration to 19.0

Sadiq-OSI:
## Summary

- Migrates `stock_picking_operations_multilang` to Odoo 19.0
- Preserves warehouse-configurable language selection for Picking Operations reports
- Adds the module to the 19.0 repository index and metapackage

## Validation

- Installed/updated the module on Odoo 19.0
- Ran the module test suite on Odoo 19.0
- Confirmed uninstall restores `stock.action_report_picking` to `stock.report_picking`
- Manually compared Odoo 18 and Odoo 19 Picking Operations reports:
  - Partner language mode prints using the partner language
  - Warehouse language mode prints using the configured warehouse language
  - Blank mode falls back to standard Odoo behavior
  - Partner mode with no partner falls back without crashing


stock_picking_operations_multilang/tests/test_stock_picking_operations_multilang.py
```diff
@@ -1,10 +1,15 @@
...
+from unittest.mock import patch
 
+from odoo import Command
 
-class TestPickingOperationsLanguage(BaseCommon):
+from odoo.addons.stock.tests.common import TestStockCommon
+from odoo.addons.stock_picking_operations_multilang.hooks import uninstall_hook
+
+
+class TestPickingOperationsLanguage(TestStockCommon):
```
kalpeshgajera-uncannycs: `class TestPickingOperationsLanguage(BaseCommon):`

Suggest using BaseCommon here instead of TransactionCase. It already disables tracking by default, so the setUpClass override redefining cls.env with tracking_disable=True becomes unnecessary. It also exposes a shared res.partner, so you don't need to create one manually. This keeps the test setup consistent with core conventions and removes duplicated boilerplate.

Sadiq-OSI: Updated, thanks. The test now inherits from BaseCommon and reuses the shared partner, with only the required stock refs set up explicitly.

PR #513, opened 2026-06-22
[19.0][MIG] stock_picking_report_qty_undelivered

Sadiq-OSI:
## Description

Migration of `stock_picking_report_qty_undelivered` to Odoo 19.0.

This module adds an option on delivery operation types to print a summary of undelivered quantities on the delivery slip when a picking is partially delivered without creating a backorder.

## Changes

- Ported the addon from 18.0 to 19.0.
- Updated the manifest version to `19.0.1.0.0`.
- Updated the picking type form view inheritance for Odoo 19.
- Updated delivery slip QWeb inheritance to match Odoo 19 report structure and quantity formatting.
- Updated tests for Odoo 19 stock move/report behavior.
- Added the addon to the root README and metapackage dependencies.

## Validation

- Installed the module successfully on Odoo 19.
- Ran the module test suite successfully:
  - `0 failed, 0 error(s)`
  - PDF smoke test was skipped locally because wkhtmltopdf is not installed.
- Ran module update successfully on Odoo 19.
- Manually tested delivery slip behavior:
  - With the option enabled, partially undelivered products appear in the summary.
  - Fully delivered products are excluded from the summary.
  - With the option disabled, the summary disappears.


stock_picking_report_qty_undelivered/tests/test_stock_picking_report_qty_undelivered.py
kalpeshgajera-uncannycs: `class TestReportQtyUndelivered(BaseCommon):`

Suggest using `BaseCommon` here instead of `TransactionCase`. It already disables tracking by default, so the `setUpClass` override redefining `cls.env` with `tracking_disable=True` becomes unnecessary. It also exposes a shared `res.partner`, so you don't need to create one manually. This keeps the test setup consistent with core conventions and removes duplicated boilerplate.

Sadiq-OSI: Thanks, updated to use BaseCommon and removed the manual partner setup.

PR #522, opened 2026-06-26
[17.0][IMP] stock_picking_report_valued: add sale.order.line values on sol_vals if sale_line._cache is empty.

ajfebres:
@binhexTeam

This change fixes an issue in stock_picking_report_valued when rendering a valued picking report during email sending.

In this flow, sale_line._cache may be incomplete, so _convert_to_write(sale_line._cache) can return sol_vals without the minimum required sale.order.line fields (for example: product_id, order_id, discount, price_unit).
When that happens, creating the virtual line and triggering amount/tax computation can fail in order_line._compute_all().

**What was changed**
A fallback was added to complete sol_vals whenever key fields are missing:

product_id
order_id
discount
price_unit

Additionally, after creating the virtual line with new(sol_vals), _compute_amount() is explicitly called to ensure amounts and taxes are properly recomputed with complete values.

**Expected result**
Sending the email no longer breaks valued picking computation when sale.order.line cache is incomplete.
The valued report keeps consistent subtotal and tax values.

### DIFF
--- a/stock_picking_report_valued/models/stock_move_line.py
+++ b/stock_picking_report_valued/models/stock_move_line.py
@@ -75,8 +75,18 @@ class StockMoveLine(models.Model):
                 # Create virtual sale line with stock move line quantity
                 sol_vals = line.sale_line._convert_to_write(line.sale_line._cache)
                 sol_vals["product_uom_qty"] = quantity
+                if not sol_vals.get("product_id", None):
+                    sol_vals.update(
+                        {
+                            "product_id": line.sale_line.product_id.id,
+                            "order_id": line.sale_line.order_id.id,
+                            "discount": line.sale_line.discount,
+                            "price_unit": line.sale_line.price_unit,
+                        }
+                    )
                 sol_vals.pop("price_subtotal", None)
                 valued_line = line.sale_line.new(sol_vals)
+                valued_line._compute_amount()
             line.update(
                 {
                     "sale_tax_description": ", ".join(


ajfebres:
@rrebollo PING!

rrebollo:
please prefix the PR title with base branch: "[17.0] ..."

rrebollo:
The PR description if fine. Can you attach a traceback of the issue? Also, would you be open to adding a test that exposes the issue, so we can avoid regressions in the future?

stock_picking_report_valued/models/stock_move_line.py
```diff
@@ -75,8 +75,18 @@ def _compute_sale_order_line_fields(self):
...
+                        {
+                            "product_id": line.sale_line.product_id.id,
+                            "order_id": line.sale_line.order_id.id,
+                            "discount": line.sale_line.discount,
+                            "price_unit": line.sale_line.price_unit,
+                        }
+                    )
                 sol_vals.pop("price_subtotal", None)
                 valued_line = line.sale_line.new(sol_vals)
+                valued_line._compute_amount()
```
rrebollo, line 89:
```suggestion
```

Are you sure this is really needed? From the description, the motivation seemed to be about safeguarding. An explicit call to something that should be ORM‑triggered feels a bit forced, and it also doesn't look very performance‑friendly.

rrebollo (commented):
Thanks for your contribution. Please consider my suggestions.

PR #531, opened 2026-08-04
[19.0][MIG] stock_quant_history: Migration to 19.0

Sadiq-OSI:
This PR migrates `stock_quant_history` from 18.0 to 19.0.

Main changes:

- Adapt Odoo 19 APIs, domains, timezone handling, views, and test fixtures.
- Add company-scoped snapshots, move-line processing, base snapshots, and record rules.
- Preserve location-derived company reporting, including shared locations.
- Prevent repeated and concurrent snapshot generation.
- Prevent changing the company of generated snapshots.
- Add a post-migration script that splits legacy mixed-company snapshots without losing history lines.
- Replace private SQL flush handling with supported flush and cache-invalidation APIs.
- Expand coverage for multi-company security, migration cases, timezones, rounding, UoMs, location usages, and cache consistency.
- Update generated documentation and repository metadata.


stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -60,41 +66,86 @@ class StockQuantHistorySnapshot(models.Model):
...
         readonly=True,
+        check_company=True,
         help="Base snapshot used to generate this snapshot",
     )
 
     @api.depends("inventory_date")
+    @api.depends_context("lang", "tz")
     def _compute_name(self):
         # Odoo enforce users to be linked to an active lang
         lang = self.env["res.lang"]._lang_get(self.env.user.lang)
```
kalpeshgajera-uncannycs: Could we use the effective environment/context language here instead of self.env.user.lang?

This method declares @api.depends_context("lang", "tz"), so the computed name is expected to react to the lang context as well. Using self.env.user.lang bypasses an explicit with_context(lang=...).

Please check whether this should use the effective environment language so the lang dependency is consistent with the implementation.

stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -108,11 +159,13 @@ def _allowed_location_usage(self):
 
     def _generate_stock_quant_history(self):
         self.ensure_one()
+        self._lock_for_generation()
         self.generated_date = fields.Datetime.now()
         previous_quant_snapshot = self.search(
             [
                 ("state", "=", "generated"),
                 ("inventory_date", "<=", self.inventory_date),
```
kalpeshgajera-uncannycs, line 159: The runtime previous-snapshot lookup uses inventory_date <= self.inventory_date, while the migration logic uses inventory_date < %s.

Could we keep the behavior consistent between normal generation and migration? In particular, should a snapshot generated for the same inventory date be considered as the previous snapshot?

If same-date snapshots are intentionally supported, please clarify the expected ordering/selection because the current order="inventory_date desc" is not deterministic for multiple generated snapshots with the same date.

stock_quant_history/models/stock_quant_history_snapshot.py
```diff
@@ -60,41 +66,86 @@ class StockQuantHistorySnapshot(models.Model):
...
+        try:
+            self._lock_and_refresh(["state"])
+        except LockError:
+            raise UserError(
+                self.env._("A selected snapshot is already being generated.")
+            ) from None
+        if any(snapshot.state != "draft" for snapshot in self):
+            raise UserError(self.env._("Only draft snapshots can be generated."))
+
+    def write(self, vals):
```
kalpeshgajera-uncannycs, line 116: The README states that generated snapshots are immutable, but the current ORM protection only prevents changing company_id.

generated_date and previous_snapshot_id are still writable through ORM/view access after generation. Could we enforce the documented immutability at the model level for the fields that are intended to be immutable once the snapshot leaves draft?

Please also verify which fields are intentionally allowed to change as part of the generation workflow, especially state, so that the protection does not interfere with the existing generation flow.

BhaveshHeliconia (approved):
Code review looks good to me!

PR #480, opened 2026-03-02
[18.0][MIG] stock_move_variation_report: Migration to 18.0

dreispt:
- [ADD] stock_move_variation_report
- [IMP] stock_move_variation_report: pre-commit auto fixes
- [MIG] stock_move_variation_report: Migration to 18.0

PR #534, opened 2026-08-17
[19.0][MIG] stock_move_variation_report: Migration to 19.0

dreispt:
## Summary

Migration of `stock_move_variation_report` to 19.0, preserving the commit history from the 18.0 migration in #480.

This module adds a `Qty Variation` computed column on `stock.move.line` and makes it visible in the Inventory / Reporting / Product Moves tree view.

Assisted-by: Devin


stock_move_variation_report/models/stock_move_line.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+
+    def _get_qty_variation_sign(self):
+        self.ensure_one()
+        if self.move_id._is_in():
+            return 1
+        elif self.move_id._is_out():
+            return -1
+        else:
+            return 0
```
BhaveshHeliconia, line 22:
```suggestion
    internal_src = self.location_id.usage in ("internal", "transit")
    internal_dest = self.location_dest_id.usage in ("internal", "transit")
    if not internal_src and internal_dest:
        return 1
    elif internal_src and not internal_dest:
        return -1
    return 0
```
_is_in() and _is_out() only return True when a line is marked as picked. Before that, the variation calculates as 0 and stays stuck at 0 in the database. Checking location types directly calculates the sign reliably right when lines are created and avoids relying on accounting methods.

stock_move_variation_report/models/stock_move_line.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @api.depends(
+        "quantity",
+        "move_id.product_uom_qty",
```
BhaveshHeliconia, line 26:
```suggestion
```
 isn't used in the calculation, so we can clean it

stock_move_variation_report/__manifest__.py
```diff
@@ -0,0 +1,13 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Stock Move Report Show Quantity Variation",
+    "summary": "Show quantity variation the the Inventory Moves report",
+    "version": "19.0.1.0.0",
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/stock-logistics-reporting",
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "depends": ["stock_account"],
```
BhaveshHeliconia, line 11:
```suggestion
    "depends": ["stock"],
```

stock_move_variation_report/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+# © 2025 Daniel Reis - Open Source Integrators
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Stock Move Report Show Quantity Variation",
+    "summary": "Show quantity variation the the Inventory Moves report",
```
BhaveshHeliconia, line 5:
```suggestion
    "summary": "Show quantity variation in the Inventory Moves report",
```

PR #464, opened 2026-01-12
[19.0][MIG] stock_picking_report_custom_description: Migration to 19.0

miguel-S73:
Standard migration to 19.0

PR #535, opened 2026-08-31
[19.0][MIG] stock_picking_report_custom_description: Migration to version 19.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/stock-logistics-reporting/pull/464
TT64146
@Tecnativa @pedrobaeza @carlosdauden @sergio-teruel @juancarlosonate-tecnativa @chienandalu @BhaveshHeliconia could you please review this?


BhaveshHeliconia (approved):
Code and functional review LGTM!

stock_picking_report_custom_description/models/stock_rule.py
```diff
@@ -41,5 +41,4 @@ def _get_stock_move_values(
                 and description_picking != line.product_id.display_name
             ):
                 res["description_picking"] = description_picking
-            res["name"] = line.name
```
pedrobaeza: Why removing this?

stock_picking_report_custom_description/tests/test_stock_picking_report_custom_description.py
```diff
@@ -13,25 +13,39 @@ def setUpClass(cls):
...
-                "type": "consu",
-                "description_sale": "Custom description",
-            }
+            {"name": "Test product", "type": "consu"}
         )
         order_form = Form(cls.env["sale.order"])
         order_form.partner_id = cls.customer
         with order_form.order_line.new() as line_form:
             line_form.product_id = cls.product
+            line_form.name = f"{cls.product.display_name}\nCustom description"
```
pedrobaeza: Why adding the display_name everywhere?

stock_picking_report_custom_description/models/stock_rule.py
carlos-lopez-tecnativa: Odoo removed this field in https://github.com/odoo/odoo/commit/facf4eba6cd0504aae949c23454c6ffa9eaa9c3f

stock_picking_report_custom_description/tests/test_stock_picking_report_custom_description.py
carlos-lopez-tecnativa: Hmm, I'm not sure. This change comes from the old PR.

I reviewed it, and Odoo now removes `display_name` if it is already included in `description_picking`: https://github.com/odoo/odoo/blob/89650a5f44b5835028dba57a9ff6fa30515bc5ec/addons/stock/models/stock_move.py#L2464

I imagine the intention was to test whether `product.display_name` is included in name. In the picking, this is removed. I'm not sure, but since this is a computed field (from the `sale.order.line`), I added the description to the result. I don't see any harm in this either. Please let me know whether we should keep this or not.

Additionally, I use `description_picking_manual` instead of description_picking because the latter is now a computed field. See: https://github.com/odoo/odoo/commit/6b2d3af64a076654e04494972acc4c42d7c54bd8

carlos-lopez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge nobump

PR #537, opened 2026-09-06
[19.0][ADD] stock_picking_report_partner_product_unitprice: new module

NICO-SOLUTIONS:
This module adds an optional unit price to the delivery slip report.

The unit price display can be configured per partner through the Render Unit Price on Delivery Slip setting in the partner's Inventory section. When enabled, the unit price from the related sales order line is displayed on the delivery slip. If a picking contains products with different unit prices, the delivery slip keeps them in separate aggregated lines. Products with the same unit price can still be aggregated.

The unit price column is only displayed when the setting is enabled and at least one delivery line has a unit price from a related sales order line.

The setting is company-dependent, allowing different configurations for the same partner across multiple companies.


stock_picking_report_partner_product_unitprice/report/report_deliveryslip.xml
```diff
@@ -0,0 +1,68 @@
...
+                t-if="render_agg_unitprice"
+            >
+                <span
+                    t-if="aggregated_lines[line].get('unit_price')"
+                    t-out="aggregated_lines[line]['unit_price']"
+                    t-options='{"widget": "monetary", "display_currency": aggregated_lines[line]["currency_id"]}'
+                />
+            </td>
+        </xpath>
+    </template>
```
BhaveshHeliconia, line 67:
```suggestion
    </template>
    <template
    id="stock_report_delivery_has_serial_move_line_inherit"
    inherit_id="stock.stock_report_delivery_has_serial_move_line"
>
    <xpath expr="//td[@name='move_line_lot_quantity']" position="after">
        <td class="text-end" name="move_line_lot_unit_price" t-if="render_unitprice">
            <span
                t-if="move_line.move_id.sale_line_id"
                t-esc="move_line.move_id.sale_line_id.price_unit"
                t-options='{"widget": "monetary", "display_currency": move_line.move_id.sale_line_id.currency_id}'
            />
        </td>
    </xpath>
</template>
```
When deliveries include lot or serial numbers, the header adds the "Unit Price" column, but the table body calls stock_report_delivery_has_serial_move_line. Without this template override, rows only have 3 cells under a 4-column header, breaking the PDF table layout.

stock_picking_report_partner_product_unitprice/tests/test_stock_move_line_unitprice.py
```diff
@@ -0,0 +1,157 @@
+# Copyright 2026 NICO SOLUTIIONS - ENGERINEERING & IT
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+
+from odoo.tests import TransactionCase
+
+
+class TestStockMoveLineUnitPrice(TransactionCase):
```
BhaveshHeliconia:
```suggestion
class TestStockMoveLineUnitPrice(Basecommon):
```

stock_picking_report_partner_product_unitprice/tests/test_stock_move_line_unitprice.py
```diff
@@ -0,0 +1,157 @@
...
+            {
+                "name": "Test Pricelist",
+                "currency_id": cls.currency.id,
+            }
+        )
+        cls.partner = cls.env["res.partner"].create(
+            {
+                "name": "Test Partner",
+            }
+        )
```
BhaveshHeliconia:
```suggestion
```

stock_picking_report_partner_product_unitprice/report/report_deliveryslip.xml
```diff
@@ -0,0 +1,68 @@
...
+        <xpath expr="//th[@name='th_sm_quantity']" position="after">
+            <th t-if="render_unitprice" name="th_sm_unit_price" class="text-end">
+                <strong>Unit Price</strong>
+            </th>
+        </xpath>
+        <xpath
+            expr="//tr[@t-foreach='lines']/td[hasclass('o_td_quantity')][last()]"
+            position="after"
+        >
+            <td class="text-end" t-if="render_unitprice and move.sale_line_id">
```
BhaveshHeliconia:
```suggestion
            <td class="text-end" t-if="render_unitprice">
```
If an order contains both sale and non-sale lines, putting and move.sale_line_id on the <td> drops the entire table cell on non-sale lines, causing table column misalignment.

stock_picking_report_partner_product_unitprice/models/stock_move_line.py
```diff
@@ -0,0 +1,21 @@
...
+        current_move = move_line.move_id if move_line else move
+
+        if current_move.sale_line_id:
+            price = current_move.sale_line_id.price_unit
+            currency = current_move.sale_line_id.currency_id
+            res["line_key"] = f"{res['line_key']}_{format(price, '.2f')}_{currency.id}"
+            res["unit_price"] = price
+            res["currency_id"] = currency
+
+        return res
```
BhaveshHeliconia:
```suggestion
        price = current_move.sale_line_id.price_unit
        currency = current_move.sale_line_id.currency_id
        res["line_key"] = f"{res['line_key']}_{str(price)}_{currency.id}"
        res["unit_price"] = price
        res["currency_id"] = currency
    return res
```
Prevents AttributeError if current_move is None. Avoids splitting same-product lines on the delivery slip when the partner does not have unit prices enabled. Uses str(price) instead of format(price, '.2f') to support unit prices with more than 2 decimal places.

stock_picking_report_partner_product_unitprice/views/res_partner.xml
```diff
@@ -0,0 +1,15 @@
...
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0) -->
+<odoo>
+    <record id="view_partner_stock_form_inherit_unitprice" model="ir.ui.view">
+        <field name="name">res.partner.stock.property.form.inherit.unitprice</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="stock.view_partner_stock_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='inventory']" position="inside">
+                <field name="picking_report_render_unitprice" />
+            </xpath>
```
BhaveshHeliconia, line 12:
```suggestion
            <xpath expr="//page[@name='sales_purchases']//group" position="inside">
                <group name="inventory_delivery_reports" string="Delivery Reports">
                    <field name="picking_report_render_unitprice" />
                </group>
            </xpath>
```

stock_picking_report_partner_product_unitprice/report/report_deliveryslip.xml
NICO-SOLUTIONS: Thanks! Exactly. When lot or serial numbers are present, Ordered is replaced by Lot/Serial Number, so the header still has the same number of columns. With the added Unit Price column, the header has 4 columns, while stock_report_delivery_has_serial_move_line only renders 3 cells. This override adds the missing Unit Price cell to keep the table layout aligned.

stock_picking_report_partner_product_unitprice/views/res_partner.xml
NICO-SOLUTIONS: The inventory group will also be extended by another module (picking_report_render_barcode, https://github.com/OCA/stock-logistics-reporting/pull/536). Moving the field to a new group under sales_purchases would change the existing layout and split related delivery report settings across different locations.
I would prefer to keep the current location. Alternatively, we could introduce a shared inventory_delivery_reports container used by both fields, but that would require changes to the base module, which I don't think is necessary for this case.

BhaveshHeliconia (approved):
Code review LGTM!

PR #539, opened 2026-09-08
[18.0][FIX] stock_picking_report_valued: convert the unit price to the operation UoM

carlosdauden:
When selling in a UoM other than the product's reference one, stock generates the move in the reference UoM, so the operation ends up in a different UoM than the sale line's. The unit price was taken raw from the sale line, unconverted, and the virtual line used for the calculation mixed the operation's quantity with the order's UoM, so the valued delivery slip printed a disproportionate unit price and subtotal (an order in dozens was valued twelvefold, and that subtotal also carried through to the picking totals). The virtual line is now built entirely in the operation's UoM, and the unit price is converted to that same UoM, matching what 19.0 already does. Adds a test covering the case, which no version had covered before.

@Tecnativa

ping @sergio-teruel @CarlosRoca13 @carlos-lopez-tecnativa


stock_picking_report_valued/tests/test_stock_picking_valued.py
```diff
@@ -180,3 +180,43 @@ def _patched_get_report_valued_total_amount(self):
...
+        sale_order = self.env["sale.order"].create(
+            {
+                "partner_id": self.partner.id,
+                "company_id": self.env.user.company_id.id,
+                "order_line": [
+                    (
+                        0,
+                        0,
+                        {
+                            "product_id": self.product.id,
```
carlos-lopez-tecnativa, line 202:
```suggestion
                "order_line": [
                    Command.create(
                        {
                            "product_id": self.product.id,
```

carlos-lopez-tecnativa (approved):
LGTM, just a non-blocking comment, but if you can improve this, that would be great.

BhaveshHeliconia (approved):

sergio-teruel (approved):

stock_picking_report_valued/tests/test_stock_picking_valued.py
carlosdauden: Changes done.

carlosdauden:
Changes done.
Ready to merge

CarlosRoca13 (approved):

PR #540, opened 2026-09-18
[19.0][MIG] delivery_line_sale_line_position: Migration to 19.0

AEstLo:
image
image
image


AEstLo:
@henrybackman @sebalix @simahawk I am tagging you as responsibles of the migration to 18.0, in case you are intested to take a look. Thanks in advance!

delivery_line_sale_line_position/readme/CREDITS.md
```diff
@@ -1,2 +0,0 @@
-The migration of this module from 16.0 to 17.0 was financially supported
```
sebalix, line 1: Is the removal of this file in the pre-commit commit accidental?

About its removal I don't remember what was the policy about it, I guess it's OK to remove it in 19.0 but it should probably be part of the migration commit.

sebalix (commented):
LG, one small remark.

delivery_line_sale_line_position/readme/CREDITS.md
AEstLo: Thanks for checking!
- According to the third point of this guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0 the file should be deleted :)
- I deleted it before the pre-commit so that the README and index is updated accordingly, but I can undo de change if you think it's wrong

